PR #2238, opened 2013-04-05
Add support for Dragonfly's url_host

### LINKED ISSUE #2049: Respect asset_host completely
Right now, many URLs are hard-coded to be relative. Monkey patches are required to make Refinery respect the asset_host setting. I would like to make Refinery respect the asset host thoroughly and completely.

mattolson:
This change fixes #2049 by adding configuration support for Dragonfly's url_host setting. If you set this to your CDN or other caching server, your Rails app won't get hit at all for redundant dragonfly requests. I realize that Rack::Cache is in place, but this still allows for a further optimization for image heavy sites.

I chose to stick with Dragonfly's existing configuration because the changes are seamless for the rest of Refinery. The url_host is automatically included by Dragonfly's url method. It also gives flexibility to use a different server for dragonfly images than for other static assets, for example: a passthrough CDN for dragonfly and an S3-based Cloudfront distribution for static assets.

Note that Dragonfly does not support a %d placeholder in the host string like Rails does, it only supports a single host. Also, the string must be a full url starting with the protocol scheme (http or https).

### DIFF
--- a/images/lib/generators/refinery/images/templates/config/initializers/refinery/images.rb.erb
+++ b/images/lib/generators/refinery/images/templates/config/initializers/refinery/images.rb.erb
@@ -39,6 +39,7 @@ Refinery::Images.configure do |config|
   # config.dragonfly_secret = <%= Refinery::Images.dragonfly_secret.inspect %>
   # If you decide to trust file extensions replace :ext below with :format
   # config.dragonfly_url_format = <%= Refinery::Images.dragonfly_url_format.inspect %>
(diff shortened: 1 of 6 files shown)


CJBrew:
@hayksaakian yeah I'm a bit mystified by this stuff. I have to admit I just perturbed the code until it worked, and I'm still unsure why app A is working nicely and app B isn't. I think the fact that app A is not using CDN and app B is, might be significant.

PR #2598, opened 2014-05-16
Page list visible when editing a page

isaacfreeman:
This shows a thin page tree at the left of the edit window. The tree gives an overview of where the current page lives in the hierarchy, and makes it quicker to edit a series of pages without having to go back to the page index between each.


parndt:
I think that unfortunately it's too cluttered

PR #2615, opened 2014-06-06
Allow page parts to use customised presenters

anitagraham:
A page part will look for a named presenter and named data.
If found it will use the presenter to present the data.

Collection_presenter is a new presenter which will present any
collection as an unordered list.

This has passed the refinerycms testing, but I haven't written tests for the new behaviour yet.


anitagraham:
Close this until after 3.0 is released

bricesanchez:
You could let it opened and tag another milestone, like 4.0

parndt:
Well, 3.0 has been released..

anitagraham:
I guess I can review it for need/currency and any recommended changes

anitagraham:
I think this could be called stale. Closing.

PR #2619, opened 2014-06-21
Extend page options

anitagraham:
This pull request is a suggestion for splitting Advanced Page Options into separate sections and allowing plugins to register a page-options template which would become one of the Page Options sections.

The changes in the page edit screen are shown in the attached images.

The initial idea was driven by my need to give refinerycms-testimonials a way to control how/which testimonials were displayed on a page. The current implementation puts the testimonials options fields into an edit tab but becoming part of page options seemed more appropriate. While looking into this I noted that SEO options are treated as a special case and just included in advanced  options. Continuing to add special cases to add to advanced options is unsustainable.

This initial implementation treats Page Options and SEO options as special cases too, but leaves the way open for them to become 'regular', registered page options. (See https://github.com/anitagraham/refinerycms/blob/extend-page-options/pages/app/helpers/refinery/admin/pages_helper.rb)

page options - all closed
+ Each of the three options can be unrolled/rolled up independently.
+ Other plugins can register templates to appear in this section of the page

page options - open
+ The standard view of the Advanced Page Options doesn't include SEO options.

page options - page and seo options open
+ Page Options and SEO options unrolled. This is the same as the current system's Advanced Page Options.

page options - seo and testimonial options open
+ Plugins can register a page options template.
+ Any combination of options can be rolled up or down.

_All current tests have passed (with changes to the text of the Advanced Page Options link)_
_Tests for the new functionality have not yet been written_


anitagraham:
Wrote a reply / reread what you wrote / deleted reply.

I have written a PR which does a similar thing. I don't think I have links to view-templates though and it is aimed more at front-end changes than back-end. 
#2615 and #2636 include some discussion of the ideas.

anitagraham:
Closing this PR as it hasn't moved forward for a long time.

PR #2636, opened 2014-07-18
Rewrite admin page edit

anitagraham:
Continuing some suggested changes for integrating plugins into the page editing process. In the process I have cleaned up `_form_page_parts.html.erb` and added helpers which are called from `pages/_form_page_parts.html.erb` which is now just a couple of lines long.

The new helpers
1. Assemble all the data for the tab/tab editors into a single object.
2. Use the object to generate the tab and editor lists. This will make it easier to use different markup for the tabs should that ever be needed. Restrained myself from using presenters.
3. Add a field to page_parts that is the name of the plugin responsible for providing a template for editing.
4. Add a field to plugins which is the name of the template to pass back to the form.
5. Small changes to js and scss to accommodate the changes.

Thus if
````
page_part.plugin = 'my-plugin'
my-plugin.edit_page_template = 'add_my_information'
````
then `add_my_information` will be the template used to render the testimonials tab on a page. 

The default template is the one currently used for html body parts: `_page_part_field`, albeit reduced to a single line, as all editing tabs are wrapped in the `page_parts_attributes` fields which were originally in `page_part_field`.

It is anticipated that plugins could use this pattern to add parts to pages and make themselves responsible for their tab. 

````ruby
Refinery::Page.class_eval do
  validates     :testimonials_count, :numericality => { :only_integer => true, :greater_than_or_equal_to=>0 }
  before_save   :add_testimonial_page_part
  protected
  def add_testimonial_page_part
    parts.build(title: 'Testimonials', position: parts.count, plugin: 'refinery-testimonials') unless part_with_title('Testimonials')
  end
end
````
The helpers use content_tag a lot. I have used Markaby in the past and then mustache/handlebars more recently for templating. Does refinery have a preference?


bricesanchez:
@anitagraham and @parndt is this PR still relevant ?

PR #2665, opened 2014-08-27
Page params for plugins

anitagraham:


anitagraham:
I'll look at the advice above to add a page config option. (I'd overlooked that before).

bricesanchez:
@anitagraham @parndt  is this PR still relevant ?

anitagraham:
No, I think we are a long way past this problem now.

Issue #2690, opened 2014-09-16
Refinery Forms installation and testing problems (Generator bug?)


CJBrew:
I've found creating Forms using the generator leads to an engine that requires some 'bodging' to get installed, and cannot easily be unit tested with rspec.

E.G creating a normal engine/extension:
```Shell
rails generate refinery:engine wibble test:string num:integer
```
then
```Shell
bundle install
rails generate refinery:wibbles
rake db:migrate
rake db:seed
cd vendor/extensions/wibbles
bundle exec rake refinery:testing:dummy_app
bundle exec rake spec
```

this happens to fail its tests, but the tests do at least run.

However
```Shell
rails generate refinery:form fibble test:string num:integer
bundle install
```
 
~~Immediately fails with~~

```Shell
Fetching additional metadata from https://rubygems.org/...
Resolving dependencies...
Bundler could not find compatible versions for gem "refinerycms-settings":
  In snapshot (Gemfile.lock):
    refinerycms-settings (2.1.1)

  In Gemfile:
    refinerycms-fibbles (>= 0) ruby depends on
      refinerycms-settings (~> 2.1.2) ruby

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

~~This can be fixed by changing the version of refinerycms-settings in the gemspec for the Engine from 2.1.2 to 2.1.1.~~

_Edit: Already fixed in refinerycms 2.1.3_
 
Continuing, these commands succeed
```Shell
rails generate refinery:fibbles
rake db:migrate
rake db:seed
```

The forms can be used at this point (actually, this example is buggy for me. Not sure why). 

However when I try to run specs in the new engine -- starting by creating a dummy app as before:

```Shell
cd vendor/extension/fibbles
bundle exec rake refinery:testing:dummy_app
```

I get an error:

```Shell
rake aborted!
Don't know how to build task 'refinery:testing:dummy_app'
(See full trace by running task with --trace)
```

Similarly, the `rake spec` command fails 

```
Don't know how to build task 'spec'
```

I think the trouble is that the generator for forms is outdated. I don't know how to fix it and would be inclined to simply create an engine and copy across the views etc from the "refinerycms-inquiries" gem as needed. But it probably should 'just work' ;-)

PR #2698, opened 2014-10-08
Generator for form is broken

CJYate:
Fix the forms generator
- Generating forms was broken as of 2.1.3 release. This commit fixes the generator although it does not add the rspec tests mentioned in https://github.com/refinery/refinerycms/issues/2690
- rspec tests to follow in a future commit

### DIFF
--- a/core/lib/generators/refinery/form/templates/app/controllers/refinery/namespace/plural_name_controller.rb.erb
+++ b/core/lib/generators/refinery/form/templates/app/controllers/refinery/namespace/plural_name_controller.rb.erb
@@ -45,7 +45,7 @@ module Refinery
     protected
 
       def find_page
-        @page = Refinery::Page.find_by_link_url('/<%= plural_name %>/new', :include => [:parts])
+        @page = Refinery::Page.find_by_link_url('/<%= plural_name %>', :include => [:parts])
       end
 
     end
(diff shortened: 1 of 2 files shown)


CJBrew:
@parndt.

Hi, -- it looks like this hasn't progressed -- and I can't find another PR on the system so I guess I didn't send one. Now that I think of it, I was waiting for some direction :-)  So, I think I should resend the pull request for the above branch. What branch at your end should I send it to (e.g. 2.1 stable  or 2.1 globalize?).

In any case it looks like much of this probably needs to be merged into Edge too.  I generated a Form just now on "Edge" and came up against the circular redirect issue that I've fixed here... Can send a PR for that too but I'd better do a local rebase and merge first...

parndt:
Can you give fresh steps for how to reproduce the circular redirect issue on edge?

CJBrew:
I'm AFK so can't double check, but this is what I did:

In a clean refinerycms checkout,

```
rake refinery:testing:dummy_app
cd spec/dummy
rails g refinery:form things title:string number:integer
bundle install
rails g refinery:things
rake db:migrate db:seed
rails s
```

Then navigate to 0.0.0.0:3000/things

IIRC this is a routing issue as the /things/new path is redirected or
mapped to /things. Something along those lines.

PR #2713, opened 2014-10-20
Choose view template during create page process

mattherick:
I added an extra step for choosing the view-template first and get all the correct page-parts immediately, while creating a new page.


parndt:
Further thoughts? Comments?

parndt:
OK, I think we're all agreed that we shouldn't have an extra step.. so thanks a lot for the work, but I'm not able to merge this.

PR #2728, opened 2014-10-27
Form generator: fix generator, add specs and correct the 'new' route via index

CJBrew:
I closed out the previous pull request -- this is now my latest branch. 

The forms had a number of problems and you couldn't generate a working one.

- Fixed a number of bugs
- Added a bunch of specs and fixed a few deprecation warnings on rspec.
- Create/delete functionality seems to be working well
- After previous discussion about the index/new route I reinstated the index action instead of a redirect route. It's by far the simplest solution -- the :index route simply redirects to :new in the controller.
- Forms can be created with all sorts of parameters and the specs (e.g. for creating an item) don't cover all the different field types that can be added to a model. 
- Spam filter feature is broken. If you turn it on via the generator option there will be ~6 failing tests and I don't know what will or won't work!

I'd be happy to work on fixing the spam filter in, but there's a few details about that I really don't understand. e.g.: How can we identify which fields are going to be used for spam filtering? The spam filter seems to have to be customised for any particular model, so perhaps in the case of the form the generator should create only the minimum setup? I don't know what that is!

Chris

### DIFF
--- a/core/lib/generators/refinery/form/form_generator.rb
+++ b/core/lib/generators/refinery/form/form_generator.rb
@@ -7,6 +7,16 @@ module Refinery
 
     include Refinery::ExtensionGeneration
 
+    class_option :include_spam,
+      :desc => 'Generate extension with spam filtering',
+      :type => :boolean,
+      :default => false,
+      :required => false
+
+    def include_spam?
+      options[:include_spam]
+    end
+
     def description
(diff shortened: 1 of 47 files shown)


parndt:
I've decided that it's not a safe idea to add this to 2-1-stable, but please see #3009 for the master branch version.

PR #2732, opened 2014-11-01
Implemented support for turbolinks, when it's present.

parndt:
To enable it, just place it in the Gemfile of your application:

``` ruby
gem 'jquery-turbolinks'
gem 'turbolinks'
```

Bugs that need fixing include: some tests fail

This does _not_ place a dependency on turbolinks, it has to be in the Gemfile!


prokopsimek:
?

simi:
Is 3.0 target for this?

parndt:
No, this can happen whenever..

parndt:
But if it works now, then great!  I'm sure there's lots of nasty edge cases, though.

sintro:
So, now it is still better to remove turbo links from refinery web-application? Is it still not supported?

bricesanchez:
@sintro Turbolinks works well on frontend with Refinery CMS, i think this PR was mainly for Backend UI.

sintro:
@bricesanchez, i found this PR because of buggy frontend exaclty with turbolinks enabled.
A bunch of Bootstrap`s javascripts were bugging until I disabled `\\=require turbolinks` from application.js.
For example:
1. Navbar collapsing. Open new browser window with some page. Make mobile-width window. Try to collapse in-out menu some times. Everything is OK. Now click on some menu item to change your page. Now menu can be opened fine, but it can not be closed (only some buggy movement for a moment). This bug is very similar to one when you include bootstrap`s javascripts twice.
2. Dropdown. Alsmost same situation. I am using dropdown-doubletap.js, and it is bugging even more: on fresh window menu dropdown works well, but after you are going to some new page, it just stops working. Not even opens.

All this problems fixed for me after removing \\=require turbolinks.
The second working fix, which I currently use, is to move 'refinery/javascripts' in **head**. That is all. I don`t even need to remove turbolinks from application.js. 
For reliability I also included gem 'jquery-turbolinks', but this is not necessary for 2nd method.

bricesanchez:
@sintro You're right, i think we should now move `<%= render "/refinery/javascripts" %>` to the head tag
https://github.com/refinery/refinerycms/blob/master/core/app/views/layouts/application.html.erb#L20

See : https://github.com/turbolinks/turbolinks/tree/676c5ba30f55ab737ae9e703b7f4aee84257e8a5#reloading-when-assets-change

sintro:
@bricesanchez, looks so.
Hm, is there no more sence in ``"data-turbolinks-track" => true`` pair? Offtopic, but why did they changed "true" to "reload"?

bricesanchez:
@parndt is it ready to be merged ?

parndt:
I am not sure, try it out!

sintro:
Just faced another problem, caused by turbolinks... 
As it written in official readme:

> Turbolinks makes following links in your web application faster. Instead of letting the browser recompile the JavaScript and CSS between each page change, it keeps the current page instance alive and replaces only the body (or parts of) and the title in the head. Think CGI vs persistent process.

So both this lines:
https://github.com/refinery/refinerycms/blob/master/core/app/views/refinery/_javascripts.html.erb#L2
https://github.com/refinery/refinerycms/blob/master/core/app/views/refinery/_head.html.erb#L10
Will not be included if you will not refresh the page (or initially will go to it by address string in browser).

Just realized this when I was browsing my website and clicked in refinery-blog post page. There was no styles, included by `<% content_for :stylesheets, stylesheet_link_tag('refinery/blog/frontend') %>
` string. When I refreshed the page, everything loads correctly. 

Probably, this can be solved easily for stylesheets by moving it to the top of body tag. 
But we should be care when trying to do the same with javascripts, because by default it will be reevaluated every time, when user repeats the page loading (not by refresh button, but with back/forward browser`s buttons or by sites links navigation), while the previous state of the page will be the same (is it not changed on server). So, for some scripts this can cause problems like duplicated elements or bugs in behaviour.

What`s your suggestions?

sintro:
Update on post https://github.com/refinery/refinerycms/pull/2732#issuecomment-234510388
Just tested this bug with new turbolinks 5.0.0. They have changed the way how it is managing **body** and **head** changes when you are browsing in website. The body still fully changes, and **head** now _merges_, it includes the stylesheets and javascript tags including or excluding. So everything I wrote in that post seems to be fixed in turbolinks. The problem is there can be a bunch of new bugs, because adapting the system for correct working with turbolinks is always the pain, and  with new version there are some big changes, so refinery's javascripts should be reviewed carefully.

parndt:
@sintro thank you very much!

bricesanchez:
I will prefer to rewrite the admin JS to be sure it will be compatible with Turbolinks in backend.

PR #2811, opened 2014-12-13
Add frontend and backend routes method

bricesanchez:
Hello all !

In following of our conversation in the goal to translate engine routes. 
I submit this PR in order to add the feature of frontend and backend routes in routes.rb

Now i can do this :

```ruby 
Refinery::Core::Engine.frontend_routes do
  # Frontend routes
  namespace :inquiries, :path => '' do
    # localized do
      get Refinery::Inquiries.page_path_new, :to => 'inquiries#new', :as => 'new_inquiry'

      resources :contact, :path => Refinery::Inquiries.post_path, :only => [:create],
                :as => :inquiries, :controller => 'inquiries'

      resources :contact, :path => '', :only => [], :as => :inquiries, :controller => 'inquiries' do
        get :thank_you, :path => Refinery::Inquiries.page_path_thank_you, :on => :collection
      end
    # end
  end
end

Refinery::Core::Engine.backend_routes do
  # Admin routes
  namespace :inquiries, :path => '' do
    namespace :admin, :path => Refinery::Core.backend_route do
      resources :inquiries, :only => [:index, :show, :destroy] do
        get :spam, :on => :collection
        get :toggle_spam, :on => :member
      end

      scope :path => 'inquiries' do
        resources :settings, :only => [:edit, :update]
      end
    end
  end
end
```

I'm really inspired of Spree : https://github.com/spree/spree/blob/b66ea0229646062acb7b6f89ac2447f5ffc862fb/core/lib/spree/core/routes.rb

It's backward compatible.

Now, i try to find a way to add the localized method of route_translator on frontend_routes. If you know how to do it, please help us! :+1:  : https://github.com/enriclluelles/route_translator/issues/75

### DIFF
--- a/core/lib/refinery/core/engine.rb
+++ b/core/lib/refinery/core/engine.rb
@@ -26,6 +26,35 @@ module Refinery
         end
       end
 
+      def self.frontend_routes(&block)
+        @frontend_routes ||= []
+        unless @frontend_routes.include?(block)
+          @frontend_routes << block
+        end
+      end
+
+      def self.backend_routes(&block)
+        @backend_routes ||= []
+        unless @backend_routes.include?(block)
+          @backend_routes << block
+        end
+      end
+
+      def self.draw_routes(&block)
+        @frontend_routes ||= []
+        @backend_routes ||= []
+        eval_block(block) if block_given?
+        @frontend_routes.each { |r| eval_block(&r) }
+        @backend_routes.each { |r| eval_block(&r) }
+        # # Clear out routes so that they aren't drawn twice.
+        @frontend_routes = []
+        @backend_routes = []
+      end
+
+      def eval_block(&block)
+        Refinery::Core::Engine.routes.eval_block(block)
+      end
+
       config.autoload_paths += %W( #{config.root}/lib )
 
       # Include the refinery controllers and helpers dynamically
--- a/core/config/routes.rb
+++ b/core/config/routes.rb
@@ -11,3 +11,5 @@ Refinery::Core::Engine.routes.draw do
 
   get '/sitemap.xml' => 'sitemap#index', :defaults => { :format => 'xml' }
 end
+
+Refinery::Core::Engine.draw_routes


core/lib/refinery/core/engine.rb
```diff
@@ -26,6 +26,35 @@ def refinery_inclusion!
...
+      def self.frontend_routes(&block)
+        @frontend_routes ||= []
+        unless @frontend_routes.include?(block)
+          @frontend_routes << block
+        end
+      end
+
+      def self.backend_routes(&block)
+        @backend_routes ||= []
+        unless @backend_routes.include?(block)
```
parndt: `if @backend_routes.exclude?(block)` reads easier.

core/lib/refinery/core/engine.rb
```diff
@@ -26,6 +26,35 @@ def refinery_inclusion!
         end
       end
 
+      def self.frontend_routes(&block)
+        @frontend_routes ||= []
+        unless @frontend_routes.include?(block)
```
parndt: `if @frontend_routes.exclude?(block)` reads easier.

Perhaps we should rename the variable from `block` which is a Ruby notion (i.e. `do .. end` is a block, but we're meaning something else.

Also does `@frontend_routes` really need to be an instance variable? We're accessing a class method `self.frontend_routes` so it's a little bit worrying that in `self.draw_routes` which is also a class method that we're using instance variables.  Perhaps I'm missing something.

parndt: To clarify, I would have thought we'd use `@@frontend_routes` or `frontend_routes` (though the latter would not stick around).

parndt:
@bricesanchez it's been 7 months, should I leave this open?

bricesanchez:
I could rebase it to master if it's still a PR we want for the future of Refinery CMS.

I don't find how to localize routes but i could re-work on it for the next release.

bricesanchez:
It has been replaces by the #3039 PR.

PR #2821, opened 2014-12-29
Redirect to root_path unless admin_root_path

developer88:
If just a user tries to navigate to admin#index it receives 'Cannot redirect to nil!'. I propose to redirect to root path.


parndt:
I have a hesitation based around the implementation.  Currently the method is called `refinery_admin_root_path` but it is, in fact, returning `refinery.root_path` under the case where they're not logged in.  This is kind of confusing!

Furthermore, I'm confused as to why this method is even being accessed when the user is not logged in, as this is an admin method. Perhaps we have another design error (i.e. we're not restricting early enough) that this would simply be covering up?

parndt:
I believe that if we are to patch this, @developer88, it should be here: https://github.com/refinery/refinerycms/blob/master/core/app/controllers/refinery/admin/core_controller.rb#L5

```ruby
module Refinery
  module Admin
    class CoreController < ::Refinery::AdminController
      def index
        redirect_to(refinery_admin_root_path || refinery_root_path)
      end
```

PR #2824, opened 2015-01-07
Add "rails" to line 28

alialjunied:
Abit misleading for a newbie to ruby on rails. I did not realise I had to append the "rails" command to this line while going through the guide.


alialjunied:
My apologies. I was too enthusiastic to add this change as I am probobaly doing something wrong while trying to install refinerycms

PR #2825, opened 2015-01-07
Create 6 - Generates a custom form.textile

bricesanchez:
Small adaptation of the USAGE file of form generator. It's simple and effective. 

It's great to have this guide because not everyone knows that this generator exists before having a deep look of the source code of RefineryCMS.

[ci skip]

### DIFF
--- a/doc/guides/4 - Refinery Extensions/6 - Generate a custom form.textile
+++ b/doc/guides/4 - Refinery Extensions/6 - Generate a custom form.textile
@@ -1,8 +1,8 @@
-h2. Generates a custom form
+h2. Generate a custom form
 
 h3. Description
 
-Generates a custom form based extension for Refinery automatically.
+Generate a custom form based extension for Refinery automatically.
 It works very similarly to the Refinery Engine generator.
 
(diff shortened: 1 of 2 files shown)


parndt:
When contributing guides, can you please put `[ci skip]` in your commit message? Normally I do this:
```
git commit -m "Create 6 - Generates a custom form.textile
Small adaptation of the USAGE file of form generator. It's simple and effective. 

It's great to have this guide because not everyone knows that this generator exists before having a deep look of the source code of RefineryCMS.

[ci skip]"
```
Note how it's nested right down the bottom.

Thanks for the contribution - the thing I'd say is that it's way out of context as a guide and words like "generates" probably should be "generate".. guides should take the form of a set of instructions to perform a task.  I get why this has happened, because it's copied from the usage docs. :smile:

bricesanchez:
Thanks for your feedback.

My goal was to provide us a quick way to show that RefineryCMS can generate forms. The best place should be a feature list but it's great to have a summarized guide instead of nothing :).

parndt:
:+1: thanks!

PR #2838, opened 2015-01-19
quiet_assets is only required in development mode

parndt:


simi:
I'm not sure why we're forcing this anyway since it doesn't work with Webrick AFAIK. But it is ok to merge.

simi:
:heart:

parndt:
Works with my Webrick

simi:
I :heart: your WEBrick (probably it is ok NOW).

Issue #2835, opened 2015-01-16
Could not find gem 'refinerycms-i18n'


CJBrew:
There seems to be something wrong with the engine generator. EITHER THAT or I'm doing something silly...

I get the following error on running `bundle install` in the extension folder:
__Could not find gem 'refinerycms-i18n (~> 3.0.0) ruby', which is required by gem
'refinerycms-things (>= 0) ruby', in any of the sources.__

Script to reproduce
```bash
rvm install 2.1.5
rvm 2.1.5@refinery_rails419 --create

gem install rails --version 4.1.9
gem install execjs

rails new TestApp -m http://refinerycms.com/t/edge
cd TestApp/

bundle install

rake db:drop db:create db:migrate db:seed

spring stop # because spring makes rails g hang...

rails g refinery:engine thing info:string

cd vendor/extensions/things
bundle install
```
Std Output:

```text
<LOTS CUT>
      create  vendor/extensions/things/tasks/rspec.rake
      create  vendor/extensions/things/tasks/testing.rake
      create  vendor/extensions/things/spec/support/factories/refinery/things.rb
      create  vendor/extensions/things/spec/features/refinery/things/admin/things_spec.rb
      create  vendor/extensions/things/spec/spec_helper.rb
      create  vendor/extensions/things/spec/models/refinery/things/thing_spec.rb
      append  Gemfile
------------------------
Now run:
bundle install
rails generate refinery:things
rake db:migrate
rake db:seed
Please restart your rails server.
------------------------
Updating https://github.com/refinery/refinerycms.git
Fetching gem metadata from https://rubygems.org/.........
Fetching additional metadata from https://rubygems.org/..
Resolving dependencies...
Could not find gem 'refinerycms-i18n (~> 3.0.0) ruby', which is required by gem
'refinerycms-things (>= 0) ruby', in any of the sources.
```

CJBrew:
Full output file https://gist.github.com/CJBrew/7e6a0468f12ae2884fca

parndt:
Fixed by releasing refinerycms-i18n 3.0.0

PR #2839, opened 2015-01-19
Add refinerycms-i18n to gemfile template for engines

CJBrew:
To fix issue https://github.com/refinery/refinerycms/issues/2835

This adds the dependency to an engine's Gemfile.

Manually tested with a dummy app.

### DIFF
--- a/core/lib/generators/refinery/engine/templates/Gemfile
+++ b/core/lib/generators/refinery/engine/templates/Gemfile
@@ -10,6 +10,8 @@ git 'https://github.com/refinery/refinerycms.git', :branch => 'master' do
   end
 end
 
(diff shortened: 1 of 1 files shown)


simi:
@parndt should be this locked to stable release once new stable release will be out?

Also @CJBrew can you use new has syntax?

CJBrew:
Tweaked the hash syntax.  Thanks.

parndt:
@simi yes absolutely, though currently it breaks out of the box so this seems good.  @CJBrew Do we need this for forms as well?

CJBrew:
@parndt  yes, I think it'll be needed on Forms. 

But I've not attempted to merge the work I did on forms generator up to Edge, so there's currently not even a Gemfile to add it to.

parndt:
No longer required because of #2846

Issue #2836, opened 2015-01-16
undefined method `refinery_login_with'


CJBrew:
On creating a new engine, it's not possible to run the RSpec tests.

I get an error of undefined method for "refinery_login_with" -- this is in the Feature spec at https://github.com/refinery/refinerycms/blob/master/core/lib/generators/refinery/engine/templates/spec/features/refinery/namespace/admin/plural_name_spec.rb.erb

It is apparently not including the helper function at refinerycms/testing/lib/refinery/testing/feature_macros/authentication.rb.

```shell
$ rails new my_new_application -m http://refinerycms.com/t/edge
<does stuff>
$ cd new_application
$ rails g refinery:engine my_item title:string info:text
<does stuff>
$ bundle install
$ rails generate refinery:my_items
$ rake db:migrate
$ rake db:seed
$ cd vendor/extensions/my_items/
$ bundle install

Fetching https://github.com/refinery/refinerycms.git
Fetching gem metadata from https://rubygems.org/.........
Fetching additional metadata from https://rubygems.org/..
Resolving dependencies...
Could not find gem 'refinerycms-i18n (~> 3.0.0) ruby', which is required by gem 'refinerycms-my_items (>= 0) ruby',
in any of the sources.
(see https://github.com/refinery/refinerycms/issues/2835)

$ echo "gem 'refinerycms-i18n', git: 'https://github.com/refinery/refinerycms-i18n', branch: 'master'" >> Gemfile 
$ bundle install
$ rspec

<ERRORS>

/home/cjbrew/t/new_application/vendor/extensions/my_items/spec/features/refinery/my_items/admin/my_items_spec.rb:8:in `block (4 levels) in <top (required)>': undefined method `refinery_login_with' for RSpec::ExampleGroups::Refinery::MyItems::Admin::MyItems:Class (NoMethodError)
...
```

 ***undefined method "refinery_login_with" for RSpec::ExampleGroups::Refinery::MyItems::Admin::MyItems:Class (NoMethodError)***

The full error output:
```
$ rspec
/home/cjbrew/t/new_application/vendor/extensions/my_items/spec/features/refinery/my_items/admin/my_items_spec.rb:8:in `block (4 levels) in <top (required)>': undefined method `refinery_login_with' for RSpec::ExampleGroups::Refinery::MyItems::Admin::MyItems:Class (NoMethodError)
[...]

PR #2842, opened 2015-01-20
Rspec issue in Engine generator

CJBrew:
Re: https://github.com/refinery/refinerycms/issues/2836

I believe I have fixed three small bugs I discovered as part of trying to resolve the above issue:
1)  commit 9487762: refinery_login_as function wasn't found: add `extend Refinery::Testing::FeatureMacros::Authenication` to the features/admin spec generator.
2) commit df6f299: add `rspec-its` gem to the engine Gemfile template
3) commit 9487762: Correct the symbol name for refinery_user for login

Now, my tests are still failing because it can't `visit refinery.some_admin_path`:

```
 Refinery Bleeees Admin bleeees destroy should succeed
     Failure/Error: visit refinery.bleeees_admin_bleeees_path
     NameError:
       undefined local variable or method `refinery' for #<RSpec::ExampleGroups::Refinery::Bleeees::Admin::Bleeees::Destroy:0x00000008268fb0>
```
Any assistance to nail this next issue down would be very welcome!
cheers,
Chris

--

NOTE:

Re: 1) I am surprised this was needed as it wasn't in previous versions. I think there is a better way. E.g., the file `core/spec/support/refinery.rb` which does the `extend` as default for all feature specs. Presumably this support file could be added to the Engine generator... I believe that or something similar would probably be a better fix. Comments?

Can you shed any light on what might have changed to break the engine specs?

### DIFF
--- a/core/lib/generators/refinery/engine/templates/Gemfile
+++ b/core/lib/generators/refinery/engine/templates/Gemfile
@@ -31,6 +31,8 @@ group :development, :test do
       gem 'therubyracer', '~> 0.11.4'
     end
   end
+
+  gem 'rspec-its'
 end
 
 # Gems used only for assets and not required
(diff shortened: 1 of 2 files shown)


parndt:
This was fixed in #2843

PR #2844, opened 2015-01-21
Clear before_inclusion_procs and after_inclusion_procs after calling them

parndt:
This prevents a bug where supplied procs were run over and over again,
resulting in e.g. plugins being registered multiple times.

### DIFF
--- a/core/lib/refinery/core/engine.rb
+++ b/core/lib/refinery/core/engine.rb
@@ -17,12 +17,12 @@ module Refinery
         # a before_inclusion and after_inclusion step that calls procs registered by the
         # Refinery::Engine#before_inclusion and Refinery::Engine#after_inclusion class methods
         def refinery_inclusion!
-          before_inclusion_procs.each(&:call)
+          before_inclusion_procs.each(&:call).clear
 
           Refinery.include_once(::ApplicationController, Refinery::ApplicationController)
           ::ApplicationController.send :helper, Refinery::Core::Engine.helpers
 
-          after_inclusion_procs.each(&:call)
+          after_inclusion_procs.each(&:call).clear
         end
       end
 


simi:
medal

parndt:
Hooray!!

parndt:
Debugging process:

* I put a `binding.pry` in the `before_inclusion` block inside `refinerycms-wymeditor` and it invoked it 3 times
* and I got it because it was an error with the plugin being inserted multiple times per user
* and so I figured that the plugin was being registered multiple times
* so I wondered why it'd call the same block 3 times
* and so I figured well hey let's clear work after we do it
* it had two more lines of code previously, then I remembered we can chain `.each().clear` as it returns an emuerable each time

bricesanchez:
Hello @parndt and @simi 

Since this PR was merged, when i change a value in a locale file, my refinerycms bug, i have 2 differents  error, one in frontend, the other in backend : 

Frontend : 
```ruby
undefined method `render_content_page' for #<#<Class:0x007ffb0a6a12f8>:0x007ffb0a6a03f8>

Extracted source (around line #1):        
<%= render_content_page(@page, {
      :hide_sections => local_assigns[:hide_sections],
      :can_use_fallback => !local_assigns[:show_empty_sections] && !local_assigns[:remove_automatic_sections]
    }) %>
<%= render '/refinery/draft_page_message' unless @page.try(:live?) -%>
```

Backend : 
```ruby
NoMethodError in Refinery::OrderingSupplies::Admin::OrderingSuppliesController#show
undefined method `refinery_user?' for #<#<Class:0x007fb983a71428>:0x007fb985aace40>

Extracted source (around line #1):
<% if refinery_user? && "#{controller_name}##{action_name}" !~ %r{preview#show} %>
  <% unless admin? # all required JS included by backend. %>
    <% content_for :stylesheets, stylesheet_link_tag('refinery/site_bar') unless !!local_assigns[:exclude_css] %>
    <%= yield(:stylesheets) unless local_assigns[:head] || local_assigns[:exclude_css] %>
  <% end -%>
  <div id='site_bar'>
```

I've to restart my server to be able to use it again :(

If i revert this commit, all is working again.

simi:
What about to use similar solution to https://github.com/parndt/decorators/commit/0ff56dac6a54c027992ff7585d65bc904dcc2f8b ? Maybe it can help us here.

bricesanchez:
Same error when i change a validator in a refinery extension model.

I hope the solution is in decorators :)

parndt:
Well, we have to clear the hooks, so we're gonna need to figure out another way to solve this.

bricesanchez:
This error still happen on edge after the update of decoractors. :(

PR #2848, opened 2015-01-22
Improve site bar CSS to be responsive without CSS framework

bricesanchez:
Remove useless mixin and var of rounded borders

Little clean in #editor_switch


simi:
@parndt I think this is ok for 3.0.0.

bricesanchez:
:heart:

PR #2849, opened 2015-01-22
Remove #page_parts_controls width

bricesanchez:
Fix overflow of this tag when there is a lot of page parts tabs.

This problem is here since we moved the section tabs to the left

Before : 
refinery-page-parts-before

After : 
refinery-page-part-after


simi:
Thanks @bricesanchez!

PR #2851, opened 2015-01-23
Re-add missing CSS for #records.tree. See https://github.com/refinery/re...

bricesanchez:
...finerycms/commit/afca0bc5d97186fcb8656f8d7caea47ddc1c0d2a?diff=unified

I've just re-add missing css, there is no clean-up.

Before screenshot : 
screenshot 2015-01-23 13 44 56 1

After screenshot : 
refinery-after-records-tree


parndt:
Thank you!

PR #2853, opened 2015-01-24
Update site_bar_helper.rb

terryliulj:
refinery.admin_root_path is more precise than Refinery::Core.backend_path


parndt:
The correct method to use is `refinery_admin_root_path`

simi:
Does it make sense to fix this @parndt?

parndt:
Not to me, really..

PR #2856, opened 2015-01-26
Fixed scss deprecation warning

parndt:
DEPRECATION WARNING: Extra .css in SCSS file is unnecessary. Rename core/app/assets/stylesheets/refinery/_mixins.css.scss to core/app/assets/stylesheets/refinery/_mixins.scss


simi:
There's still a lot of P`.css.scss` files around.

*Be careful, sprocket's `require` doesn't work in `scss` AFAIK.*

parndt:
So.. what should we do? Should we rename these files, or not? Or just rename them and stop using `require`?

simi:
I think rename and using just `@import` should be solution here, but it needs to be tested first.

parndt:
Fixed by #2892

PR #2857, opened 2015-01-26
Added better support for `:at` Rails mount option.

### LINKED ISSUE #2740: Bug in "Home" page button when mounting to a directory other than root
I've been deploying Refinery CMS 2.1.4 and I found a bug when I try to change it to "/subfolder" instead of "/". I follow the steps from your guide 5, changing the "mount" line in config/routes.rb file to "/subfolder" and everything seems to work. But in the "new application root" I click on "Home", it redirects me to the old root, where there is nothing, giving a Routing Error.
I add some screenshots
home - company name - mozilla firefox_002
action controller exception caught - mozilla firefox_006
Then I tried to change the link propertie for the Home page. Now the link is correct but I get a 404 error clicking on it or simply accessing to the main url.
home - page not found - company name - mozilla firefox_007
I also tried to overwrite the root rule in routes.rb with a custom one without success.
Could you say me if I'm doing something wrong?
Thank you very much.

parndt:
i.e. 
```ruby
mount Refinery::Core::Engine, at: '/some/subfolder'
```

Fixes #2740

### DIFF
--- a/core/lib/generators/refinery/cms/cms_generator.rb
+++ b/core/lib/generators/refinery/cms/cms_generator.rb
@@ -219,10 +219,11 @@ gem 'pg'
         mount = %Q{
   # This line mounts Refinery's routes at the root of your application.
   # This means, any requests to the root URL of your application will go to Refinery::PagesController#home.
-  # If you would like to change where this extension is mounted, simply change the :at option to something different.
+  # If you would like to change where this extension is mounted, simply change the
+  # configuration option `mounted_path` to something different in config/initializers/refinery/core.rb
   #
   # We ask that you don't use the :as option here, as Refinery relies on it being the default of "refinery"
-  mount Refinery::Core::Engine, at: '/'
+  mount Refinery::Core::Engine, at: Refinery::Core.mounted_path
 
 }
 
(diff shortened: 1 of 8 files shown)


simi:
I think this is good for now, but it is not a 100% Rails way.

parndt:
What would be the 100% Rails way?

simi:
IMHO make Frontend and Backend as separated engines and allow to mount them separately when needed or mount "meta-engine" for both at once.

But as I wrote before, this is ok for now.

parndt:
Ah, yes, well @bricesanchez is working on making it better over at #2811 :smile:

core/lib/generators/refinery/cms/cms_generator.rb
```diff
@@ -219,10 +219,11 @@ def mount!
         mount = %Q{
   # This line mounts Refinery's routes at the root of your application.
   # This means, any requests to the root URL of your application will go to Refinery::PagesController#home.
-  # If you would like to change where this extension is mounted, simply change the :at option to something different.
+  # If you would like to change where this extension is mounted, simply change the
+  # configuration option `mounted_path` to something different in config/initializers/refinery/core.rb
   #
   # We ask that you don't use the :as option here, as Refinery relies on it being the default of "refinery"
```
robyurkowski: Is this still the case?

parndt: Yes, unless you can find a way to work around it. We reference `refinery.path_helper` everywhere.

PR #2862, opened 2015-01-27
Image spec part4

anitagraham:
This PR rewrites the image specs to use shared examples and shared contexts.
The two contexts are the Admin Images Tab and the Add Image tab called from the page visual editor.


parndt:
@anitagraham does this replace #2812 or should I merge one and then the other?

anitagraham:
Yes, I think it does.  All I know is that git got gnotted again, and I started a new branch in desperation.

PR #2867, opened 2015-01-28
Use Capybara.default_wait_time = 10 to avoid failing tests.

simi:


parndt:
This didn't work, huh?

PR #2875, opened 2015-01-30
Custom page part title

bricesanchez:
It's a working progress. This feature works but tests are failing on my computer.

company_name_-_refinery-page-part-slug

I create this PR to share my progress and ask for inputs. I don't know if i'm doing the right things on controllers and models.

What do you think about that ?

### DIFF
--- a/core/app/assets/javascripts/refinery/admin.js.erb
+++ b/core/app/assets/javascripts/refinery/admin.js.erb
@@ -576,13 +576,17 @@ var page_options = {
   },
 
   page_part_dialog: function(){
+    var slugOptions = {
+      separator: '_'
+    }
+
     $('#new_page_part_dialog').dialog({
       title: 'Create Content Section',
       modal: true,
       resizable: false,
       autoOpen: false,
       width: 600,
-      height: 200
+      height: 300
     });
 
     $('#add_page_part').click(function(e){
@@ -590,19 +594,30 @@ var page_options = {
       $('#new_page_part_dialog').dialog('open');
     });
 
+    $('#new_page_part_title').on('input propertychange', function() {
+      var part_title = $(this).val();
+
+      if(part_title.length > 0) {
+        var part_slug = getSlug(part_title, slugOptions);
+        $('#new_page_part_slug').val(part_slug);
+      }
+    });
+
     $('#new_page_part_save').click(function(e){
       e.preventDefault();
 
       var part_title = $('#new_page_part_title').val();
+      var part_slug = $('#new_page_part_slug').val();
 
-      if(part_title.length > 0){
-        var tab_title = part_title.toLowerCase().replace(" ", "_");
+      if(part_slug.length > 0){
+        var tab_slug = getSlug(part_slug, slugOptions);
 
-        if ($('#page_part_' + tab_title).size() === 0) {
+        if ($('#page_part_' + tab_slug).size() === 0) {
           $.get(page_options.new_part_url, {
-              title: part_title
-              , part_index: $('#new_page_part_index').val()
-              , body: ''
+              title: part_title,
+              part_index: $('#new_page_part_index').val(),
+              slug: tab_slug,
+              body: ''
             }, function(data, status){
               $('#submit_continue_button').remove();
               // Add a new tab for the new content section.
@@ -621,15 +636,16 @@ var page_options = {
               // Wipe the title and increment the index counter by one.
               $('#new_page_part_index').val(parseInt($('#new_page_part_index').val(), 10) + 1);
               $('#new_page_part_title').val('');
+              $('#new_page_part_slug').val('');
 
               $('#new_page_part_dialog').dialog('close');
             }, 'html'
           );
         }else{
-          alert("A content section with that title already exists, please choose another.");
+          alert("A content section with that slug already exists, please choose another.");
         }
       }else{
-        alert("You have not entered a title for the content section, please enter one.");
+        alert("You have not entered a slug for the content section, please enter one.");
       }
     });
 
@@ -637,6 +653,7 @@ var page_options = {
       e.preventDefault();
       $('#new_page_part_dialog').dialog('close');
       $('#new_page_part_title').val('');
+      $('#new_page_part_slug').val('');
     });
 
     $('#delete_page_part').click(function(e){
--- a/pages/app/controllers/refinery/admin/page_parts_controller.rb
+++ b/pages/app/controllers/refinery/admin/page_parts_controller.rb
@@ -4,7 +4,7 @@ module Refinery
 
       def new
         render :partial => '/refinery/admin/pages/page_part_field', :locals => {
-                 :part => ::Refinery::PagePart.new(:title => params[:title], :body => params[:body]),
+                 :part => ::Refinery::PagePart.new(new_page_part_params),
                  :new_part => true,
                  :part_index => params[:part_index]
                }
@@ -20,6 +20,11 @@ module Refinery
           render :text => "'#{part.title}' not deleted."
         end
       end
+
+      protected
+        def new_page_part_params
+          params.permit(:title, :slug, :body)
+        end
 
     end
   end
--- a/pages/app/controllers/refinery/admin/pages_controller.rb
+++ b/pages/app/controllers/refinery/admin/pages_controller.rb
@@ -12,7 +12,7 @@ module Refinery
       def new
         @page = Page.new(new_page_params)
         Pages.default_parts_for(@page).each_with_index do |page_part, index|
-          @page.parts << PagePart.new(:title => page_part, :position => index)
+          @page.parts << PagePart.new(:title => page_part, :slug => page_part, :position => index)
         end
       end
 
@@ -102,7 +102,7 @@ module Refinery
         [
           :browser_title, :draft, :link_url, :menu_title, :meta_description,
           :parent_id, :skip_to_first_child, :show_in_menu, :title, :view_template,
-          :layout_template, :custom_slug, parts_attributes: [:id, :title, :body, :position]
+          :layout_template, :custom_slug, parts_attributes: [:id, :title, :slug, :body, :position]
         ]
       end
 
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -291,8 +291,12 @@ module Refinery
     #    ::Refinery::Page.first.content_for(:body)
     #
     # Will return the body page part of the first page.
-    def content_for(part_title)
-      part_with_title(part_title).try(:body)
+    def content_for(part_slug)
+      begin
+        part_with_slug(part_slug).try(:body)
+      rescue
+        raise ActiveRecord::RecordNotFound
+      end
     end
 
     # Accessor method to test whether a page part
@@ -302,22 +306,27 @@ module Refinery
     #   ::Refinery::Page.first.content_for?(:body)
     #
     # Will return true if the page has a body page part and it is not blank.
-    def content_for?(part_title)
-      content_for(part_title).present?
+    def content_for?(part_slug)
+      content_for(part_slug).present?
     end
 
     # Accessor method to get a page part object from a page.
     # Example:
     #
-    #    ::Refinery::Page.first.part_with_title(:body)
+    #    ::Refinery::Page.first.part_with_slug(:body)
     #
-    # Will return the Refinery::PagePart object with that title using the first page.
-    def part_with_title(part_title)
+    # Will return the Refinery::PagePart object with that slug using the first page.
+    def part_with_slug(part_slug)
       # self.parts is usually already eager loaded so we can now just grab
       # the first element matching the title we specified.
       self.parts.detect do |part|
-        part.title_matches?(part_title)
+        part.slug_matches?(part_slug)
       end
+    end
+
+    def part_with_title(part_title)
+      Refinery.deprecate("Refinery::Page#part_with_title", when: "3.1", replacement: "part_with_slug")
+      part_with_slug(part_title.to_s.parameterize.underscore)
     end
 
     # Protects generated slugs from title if they are in the list of reserved words
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -292,11 +292,7 @@ module Refinery
     #
     # Will return the body page part of the first page.
     def content_for(part_slug)
-      begin
-        part_with_slug(part_slug).try(:body)
-      rescue
-        raise ActiveRecord::RecordNotFound
-      end
+      part_with_slug(part_slug).try(:body)
     end
 
     # Accessor method to test whether a page part
--- a/pages/app/models/refinery/page_part.rb
+++ b/pages/app/models/refinery/page_part.rb
@@ -3,13 +3,16 @@ module Refinery
 
     belongs_to :page, :foreign_key => :refinery_page_id
 
-    validates :title, :presence => true, :uniqueness => {:scope => :refinery_page_id}
+    before_validation :set_default_slug
+
+    validates :title, :presence => true
+    validates :slug, :presence => true, :uniqueness => {:scope => :refinery_page_id}
     alias_attribute :content, :body
 
     translates :body
 
     def to_param
-      "page_part_#{title.downcase.gsub(/\W/, '_')}"
+      "page_part_#{slug.downcase.gsub(/\W/, '_')}"
     end
 
     def body=(value)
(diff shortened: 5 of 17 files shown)


core/app/assets/javascripts/refinery/admin.js.erb
```diff
@@ -591,15 +598,23 @@ var page_options = {
       $('#new_page_part_dialog').dialog('open');
     });
 
+    $('#new_page_part_title').on('keyup', function() {
```
simi: This should be on "input" event. But it doesn't work for IE8 and IE9. But alternatively "propertychange" event can be used there.

pages/app/controllers/refinery/admin/page_parts_controller.rb
```diff
@@ -4,7 +4,7 @@ class PagePartsController < ::Refinery::AdminController
 
       def new
         render :partial => '/refinery/admin/pages/page_part_field', :locals => {
-                 :part => ::Refinery::PagePart.new(:title => params[:title], :body => params[:body]),
+                 :part => ::Refinery::PagePart.new(:title => params[:title], :slug => params[:title], :body => params[:body]),
```
simi: ```:slug => params[:slug]``` :question:

parndt: params.slice(:title, :slug, :body) 👍=

parndt: Instead we should use this:

```ruby
:part => ::Refinery::PagePart.new(params.slice(:title, :slug, :body)
```
```

**Comment by bricesanchez** [comment]
a98c0d74d7e1b8163a57c9ada685d9e4 1

Hi @parndt,

I've updated this PR, could you review it and help me to understand why i've tests who fails?

**Comment by parndt** [comment]
@bricesanchez because slug is required on PagePart but the test isn't setting it.. shouldn't the slug be able to default to `title.to_param`?

**Comment by parndt** [comment]
Here's the [validator](https://github.com/bricesanchez/refinerycms/blob/page-part-title/pages/app/models/refinery/page_part.rb#L7) and here's [where it's needed](https://github.com/bricesanchez/refinerycms/blob/page-part-title/images/spec/support/shared%20contexts/visual_editor_add_image.rb#L7)

**Comment by bricesanchez** [comment]
Thanks @parndt, it works now!

> shouldn't the slug be able to default to title.to_param?
it should but how ?

**Comment by parndt** [comment]
Something like this:

```ruby
class PagePart
  def slug
    super || title.to_s.to_param
  end
end
```

bricesanchez:
Yes, i would like to use underscores instead of dashes. 

My method looks like this now :
```ruby 
class PagePart
 def slug
      super || title.to_s.parameterize
    end
end
```

But with this i'm able to save a record with a nil slug

parndt:
That's weird, is the title blank?

bricesanchez:
Nope : 

```ruby 
<Refinery::PagePart id: 10, refinery_page_id: nil, slug: nil, body: nil, position: nil, created_at: "2015-03-16 02:48:58", updated_at: "2015-03-16 02:48:58", title: "er rew e 24 ^">
```

parndt:
What about

```ruby
class PagePart
  def slug
    super.presence || title.to_s.parameterize.underscore.presence
  end
end
```

bricesanchez:
```ruby 
irb(main):001:0> p = Refinery::PagePart.new
=> #<Refinery::PagePart id: nil, refinery_page_id: nil, slug: nil, body: nil, position: nil, created_at: nil, updated_at: nil, title: nil>
irb(main):002:0> p.title = "esfefse f33\U+FFC3\U+FFA8"
=> "esfefse f33"
irb(main):003:0> p.save
   (0.1ms)  begin transaction
  Refinery::PagePart Exists (0.1ms)  SELECT  1 AS one FROM "refinery_page_parts"  WHERE ("refinery_page_parts"."slug" = 'esfefse_f33' AND "refinery_page_parts"."refinery_page_id" IS NULL) LIMIT 1
  SQL (0.2ms)  INSERT INTO "refinery_page_parts" ("created_at", "title", "updated_at") VALUES (?, ?, ?)  [["created_at", "2015-03-16 03:05:28.025965"], ["title", "esfefse f33"], ["updated_at", "2015-03-16 03:05:28.025965"]]
   (4.4ms)  commit transaction
=> true
irb(main):004:0> p = Refinery::PagePart.last
  Refinery::PagePart Load (0.1ms)  SELECT  "refinery_page_parts".* FROM "refinery_page_parts"   ORDER BY "refinery_page_parts"."id" DESC LIMIT 1
  Refinery::PagePart::Translation Load (0.1ms)  SELECT "refinery_page_part_translations".* FROM "refinery_page_part_translations"  WHERE "refinery_page_part_translations"."refinery_page_part_id" = ?  [["refinery_page_part_id", 11]]
=> #<Refinery::PagePart id: 11, refinery_page_id: nil, slug: nil, body: nil, position: nil, created_at: "2015-03-16 03:05:28", updated_at: "2015-03-16 03:05:28", title: "esfefse f33">
```

parndt:
Or probably more accurately:

```ruby
class PagePart
  before_validation :set_default_slug

  private
  def set_default_slug
    self.slug = title.to_s.parameterize.underscore.presence if self.slug.blank?
  end
end
```

bricesanchez:
Now it works well and i understand the syntax :smile:

pages/app/models/refinery/page.rb
```diff
@@ -312,11 +312,11 @@ def content_for?(part_title)
     #    ::Refinery::Page.first.part_with_title(:body)
     #
     # Will return the Refinery::PagePart object with that title using the first page.
-    def part_with_title(part_title)
+    def part_with_slug(part_slug)
```
parndt: Can you please deprecate the old API, and update the documentation above this method to be correct?
Same thing needs to happen for `content_for`,  and `content_for?`

pages/app/models/refinery/page_part.rb
```diff
@@ -18,10 +21,10 @@ def body=(value)
       normalise_text_fields
     end
 
-    def title_matches?(other_title)
-      title.present? and # protecting against the problem that occurs when have nil title
-        title == other_title.to_s or
-        parameterized_title == parameterize(other_title.to_s)
+    def slug_matches?(other_slug)
```
parndt: Can you please deprecate the old API?

pages/spec/features/refinery/admin/pages_spec.rb
```diff
@@ -646,9 +646,9 @@ module Admin
         let!(:some_page) { Page.create! :title => "Some Page" }
 
         before do
-          some_page.parts.create! :title => "First Part", :position => 1
-          some_page.parts.create! :title => "Second Part", :position => 2
-          some_page.parts.create! :title => "Third Part", :position => 3
+          some_page.parts.create! :title => "First Part",     :slug => "first_part",      :position => 1
+          some_page.parts.create! :title => "Second Part", :slug => "second_part", :position => 2
+          some_page.parts.create! :title => "Third Part",    :slug => "third_part",    :position => 3
```
parndt: this indentation seems a bit weird :smile:

parndt:
Normally we handle deprecations of public methods with the `Refinery.deprecate` method. You should find examples in the codebase already for things that are changing from 2.1 -> 3.0

It also means you're going to have to handle the old API calls until e.g. version 3.1 so for `part_with_title` that would look something like this (so that it actually calls the new API underneath with the right argument):

```ruby
def part_with_title(part_title)
  Refinery.deprecate("Refinery::Page#part_with_title", when: "3.1", replacement: "part_with_slug")
  part_with_slug(part_title.to_s.parameterize.underscore)
end
```

Understand what I mean?

bricesanchez:
Yep ! i will do that  ASAP. :smile:

bricesanchez:
output_ebnhpj

parndt:
@simi are you happy with us maintaining this feature for the rest of time?

simi:
Yes, I'm happy to maintain this for the rest of my poor live. But do we really need https://github.com/slav123/jquery-seo-url/blob/master/jquery.seourl.js included? If yes, why minificated?

bricesanchez:
Seo url support all accents like (éèàçùî....), i've changed it for the unminificated version

core/app/assets/javascripts/refinery/jquery.seourl.js
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * Dual licensed under the MIT and GPL licenses:
+ *   http://www.opensource.org/licenses/mit-license.php
+ *   http://www.gnu.org/licenses/gpl.html
+ *
+ */
+
+(function( $ ){
+
+  // protorype, and return string
```
simi: prototype (patch sent - https://github.com/slav123/jquery-seo-url/pull/1)

core/app/assets/javascripts/refinery/jquery.seourl.js
```diff
@@ -0,0 +1,282 @@
...
+  var text = '';
+
+  function trans (text) {
+    text = strtr(text, from, to)
+    text = strtr(text, special);
+    return text;
+  }
+
+  var from = 'ąĄęĘóÓśŚłŁżŻźŹćĆńŃ' + // polish
+         'ŠŒŽšœžŸ¥µÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÔÕÖØÙÚÛÜÝßàáâãäåæçèéêëìíîïðñòóôõöøùúûüýÿ'; // common
```
simi: :exclamation: missing some czech characters - č,ď,ň,ř,š,ť for example :exclamation:

I'm not sure if this is bulletproof library to generate slugs.

What about to use https://github.com/pid/speakingurl? I can create a gem for sprockets (like speakingurl-rails for example) to avoid adding this as vendored dependency since it looks like it is updating frequently.

bricesanchez:
I like your idea @simi, i would like to use a good js lib with international characters (not just english characters :P)

simi:
feel free to use this in Gemfile for now:

```ruby
gem 'speakingurl-rails', github: 'simi/speakingurl', branch: 'support-rails-assets'
```

I'll try to push this to upstream or release custom wrapper.

simi:
It is released as https://rubygems.org/gems/speakingurl-rails/versions/1.1.2 now.

bricesanchez:
I've made the changes :)

pages/app/models/refinery/page_part.rb
```diff
@@ -34,11 +42,16 @@ def normalise_text_fields
 
     private
     def parameterize(string)
+
```
simi: :scissors:

simi:
What about to use new hash syntax?

bricesanchez:
@simi, it will be great to use new hash syntax but may be in another cleaning PR.

pages/app/models/refinery/page_part.rb
bricesanchez: i will cut that, thanks.

pages/app/views/refinery/admin/pages/_form_new_page_parts.html.erb
```diff
@@ -10,3 +14,7 @@
              :cancel_button_id => "new_page_part_cancel",
              :hide_delete => true %>
 </div>
+
+<% content_for :after_javascript_libraries do %>
+  <%= javascript_include_tag 'refinery/jquery.seourl' %>
```
parndt: Where does this file come from? I don't see it in the changeset.

bricesanchez: Sorry i've forgot to remove it since we doesn't use seourl, replaced by speaking-url

parndt: Do we need to add this dependency to refinerycms-core? Shouldn't it be a dependency of refinerycms-pages?

core/app/assets/javascripts/refinery/refinery.js.erb
```diff
@@ -8,6 +8,7 @@
  *= require jquery/jquery.html5-placeholder-shim
  *= require jquery/jquery.timers
  *= require jquery/jquery.browser
+ *= require speakingurl
```
parndt: Core refinery shouldn't depend on this, it's only used for pages.

core/lib/refinery/core.rb
```diff
@@ -22,5 +22,6 @@ def root
 require 'decorators'
 require 'jquery-rails'
 require 'jquery-ui-rails'
+require 'speakingurl-rails'
```
parndt: move to pages/lib/refinery/pages.rb

core/lib/refinery/core/engine.rb
```diff
@@ -85,6 +85,7 @@ def refinery_inclusion!
           refinery/*
           refinery/icons/*
           modernizr-min.js
+          speakingurl
```
parndt: move to pages/lib/refinery/pages/engine.rb

core/refinerycms-core.gemspec
```diff
@@ -33,5 +33,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'coffee-rails',                ['~> 4.0', '>= 4.0.0']
   s.add_dependency 'jquery-rails',                '>= 2.3.0'
   s.add_dependency 'jquery-ui-rails',             '~> 5.0.0'
+  s.add_dependency 'speakingurl-rails',           '~> 1.1.2'
```
parndt: move to pages/refinerycms-pages.gemspec

pages/app/views/refinery/admin/pages/_form_new_page_parts.html.erb
bricesanchez: good point, i move it to pages

pages/app/controllers/refinery/admin/page_parts_controller.rb
```diff
@@ -4,7 +4,7 @@ class PagePartsController < ::Refinery::AdminController
 
       def new
         render :partial => '/refinery/admin/pages/page_part_field', :locals => {
-                 :part => ::Refinery::PagePart.new(:title => params[:title], :body => params[:body]),
+                 :part => ::Refinery::PagePart.new(:title => params[:title], :slug => params[:slug], :body => params[:body]),
```
parndt: I think now this could just be:

```ruby
:part => ::Refinery::PagePart.new(params.slice(:title, :slug, :body)),
```

pages/app/controllers/refinery/admin/pages_controller.rb
```diff
@@ -13,7 +13,7 @@ class PagesController < Refinery::AdminController
       def new
         @page = Page.new(new_page_params)
         Pages.default_parts_for(@page).each_with_index do |page_part, index|
-          @page.parts << PagePart.new(:title => page_part, :position => index)
+          @page.parts << PagePart.new(:title => page_part, :slug => page_part, :position => index)
```
parndt: I'm surprised this works with mass assignment... @simi?

pages/app/models/refinery/page.rb
```diff
@@ -291,8 +291,13 @@ def to_refinery_menu_item
     #    ::Refinery::Page.first.content_for(:body)
     #
     # Will return the body page part of the first page.
-    def content_for(part_title)
-      part_with_title(part_title).try(:body)
+    def content_for(part_slug)
+      begin
+        part_with_slug(part_slug).try(:body)
+      rescue
```
parndt: ..rescue _what_ exactly?  Should this only rescue `ActiveRecord::RecordNotFound`?

pages/app/models/refinery/page_part.rb
```diff
@@ -18,10 +21,15 @@ def body=(value)
       normalise_text_fields
     end
 
+    def slug_matches?(other_slug)
+      slug.present? and # protecting against the problem that occurs when have nil slug
+        slug == other_slug.to_s or
+        parameterized_slug == parameterize(other_slug.to_s)
```
parndt: let's take the opportunity to switch this to better operators:

```ruby
slug.present? && (# protecting against the problem that occurs when have two nil slugs
  slug == other_slug.to_s ||
  parameterized_slug == parameterize(other_slug.to_s)
)
```

core/app/assets/javascripts/refinery/admin.js.erb
```diff
@@ -577,13 +577,17 @@ var page_options = {
   },
 
   page_part_dialog: function(){
+    var slugOptions = {
+      separator: '_'
```
simi: Isn't `-` nicer?

ghoppe: Agree. If nothing else, Google recommends hyphens over underscores. It used to be better for SEO although I'm not sure that's still the case. 

Plus, it's marginally easier to type, matches domain restrictions, so it looks more URL-y (underscores are not allowed in domain names.)

bricesanchez: Sorry, i'm not agreed. 
We don't speak about the page slug but the page part slug.

The slug for the page part is never displayed as an entire page so Google will never see it.
 
When we call a page part in an html view, we write something like that : 
`<%= raw @page.content_for(:side_body) %>`
not
`<%= raw @page.content_for(:"side-body") %>`

parndt: Yeah this is for programmatical reference, nothing shown anywhere. I agree with @bricesanchez

pages/app/models/refinery/page.rb
parndt: still curious, this is my only objection with this PR

bricesanchez: .... Yes but i don't know why, i forget the answer. If i remove it do you merge it?

parndt: provided it works, sure!

parndt:
Thanks @bricesanchez

PR #2878, opened 2015-02-05
Generate strong params for engine admin controller.

### LINKED ISSUE #2753: Refinery 3 generator does not generate method for params
Refinery 3 generator does not generate method for params in Rails 4.

E.g.:
```ruby
def daily_item_params
  params.require(:daily_item).permit!
end
```

simi:
Closes #2753.

For

`bin/rails g refinery:engine ryba maslo banan salam`

```ruby
module Refinery
  module Rybas
    module Admin
      class RybasController < ::Refinery::AdminController

        crudify :'refinery/rybas/ryba',
                :title_attribute => 'maslo'

        private

        # Only allow a trusted parameter "white list" through.
        def ryba_params
          params.require(:ryba).permit(:maslo, :banan, :salam)
        end
      end
    end
  end
end
```

Tested in dummy app and it works! Copied directly from [rails](https://github.com/rails/rails/blob/4b0c8a9467008573c3667641ff533fe611e3df6c/railties/lib/rails/generators/rails/scaffold_controller/templates/controller.rb#L59-L66).


parndt:
Could we, and should we, test this?

simi:
We're testing only tree structure of generated folders.

Issue #2879, opened 2015-02-05
undefined method `refinery_user?' for development on various file changes


simi:
info in https://github.com/refinery/refinerycms/pull/2844#issuecomment-71684338

PR #2883, opened 2015-02-09
Add conditionnal clear on before and after inclusion procs

bricesanchez:
Fix for #2879 and #2844


parndt:
Interesting solution! 
What do you reckon, @simi ?

simi:
Actually that's based on my idea. But I'm not sure if this is enough and covers all we need.

bricesanchez:
Go merge go merge!
Give me a M, an E, a R, a G and E!!!
MERGE!! Youhou!

PR #2890, opened 2015-02-16
Fixed missing jquery-ui/datepicker dependency

hx:
Silences that pesky `undefined is not a function` message in the back end. Also, restores date picker functionality wherever it's needed!

### DIFF
--- a/core/app/assets/javascripts/refinery/refinery.js.erb
+++ b/core/app/assets/javascripts/refinery/refinery.js.erb
@@ -4,6 +4,7 @@
  *= require jquery-ui/dialog
  *= require jquery-ui/sortable
  *= require jquery-ui/tabs
+ *= require jquery-ui/datepicker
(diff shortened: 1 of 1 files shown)


parndt:
Thanks! Where do we use the datepicker?

hx:
Because I'm inheriting the project, I'm assuming it's in a file that's generated by a refinery generator. In this particular project, it's in `app/assets/javascripts/refinery/export.js`, and is only used when in the admin section.

I might be barking up the wrong tree. If so, apologies. jQuery date pickers a a bit fugly anyway ;)

parndt:
I think that it's in a custom JS file (we've never had `export.js`) and it's not requiring its dependencies properly.

A while ago we stopped requiring all of jQuery UI and just the parts that Refinery needs. I'd recommend moving the datepicker requirement to your own file that uses it :smile:

Thanks!!

PR #2893, opened 2015-02-16
Remove source of factory duplication

anitagraham:
Factory_girl_rails was loading factories, and further down in this file
so was refinery.


parndt:
I didn't have time to finish investigating.. did you check whether this loads all of your factories for you just fine?

anitagraham:
Yes, I am using it now and getting *real* errors!

PR #2895, opened 2015-02-17
Removed hard dependency on Bundler.. not sure it's required.

parndt:


simi:
Actually it doesn't work AFAIK to specify bundler version.

PR #2898, opened 2015-02-18
Spruced up the 'ol thing.

parndt:
Not ready yet..

Currently:

THE SPRUCE GOOSE?!?

This is loosely based on this UI design:

ui_15

# Help required:
Need to figure out a way to extend the white background to the bottom of the browser without showing a scroll bar or setting a background image on the body etc. Anyone know how that works best in 2015?

### DIFF
--- a/core/app/assets/stylesheets/refinery/layout.css.scss
+++ b/core/app/assets/stylesheets/refinery/layout.css.scss
@@ -1196,6 +1196,7 @@ body.dialog #content .search_form {
   margin: 0px 10px;
 }
 #dialog_container, .dialog_container {
+  margin: 0px;
   padding: 12px;
 }
 ul#image_grid, .pagination_container > ul#image_grid {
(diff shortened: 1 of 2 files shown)


parndt:
Apparently [height: 100vh](https://developer.mozilla.org/en/docs/Web/CSS/length) might be our friend - thanks @lengarvey

roonie:
Setting the min-height to 100% can do the trick. For it to work, the parent has to have a height of 100%. Using a min-height has the pro of extending when the content grows.

roonie:
Here is a quick preview how to do it
```html
  height: 100%
body
  height: 100%
  .thing_wrapper
    height: 100%
    padding: 52px 0 0
    box-sizing: padding-box
    .white_thing
      min-height: 100%
```

parndt:
@roonie nice! Thank you for the example code! :heart:

parndt:
Thanks, @roonie and @johanb for the push over the edge!

Issue #2899, opened 2015-02-18
Help: Overriding views from my private engine.


prokopsimek:
I have my private gem called "refinerycms-skeleton". It is something like refinerycms-core, but a little bit modified. 
I have there modified initializers and so on... 
I created my own edge file, which I use to generate my application.
I would like to override my gem's modified view with "rake refinery:override", but it is still overriding from refinerycms

It does mean e.g.:
1) refinerycms = app/views/refinery/_head.html.erb
2) my own gem "skeleton" = app/views/refinery/_head.html.erb
3) run "rake refinery:override" > copied view from "skeleton", not from refinerycms!
but if
1) refinerycms = app/views/refinery/_footer.html.erb
2) my own gem "skeleton" = does not have _footer.html.erb
3) run "rake refinery:override" > copied view from "refinerycms"!

My project is loading gems this way:
1) Have I my own _head view? No!
so next step 2) Has gem "refinerycms-skeleton" views _head? No!
so next step 3) So load view _head from refinerycms-core! Yes!

@parndt Could you help me how to get it? :) Is it needed to modify anything there: https://github.com/refinery/refinerycms/blob/master/core/lib/refinery/cli.rb#L113?

simi:
Look at
https://github.com/refinery/refinerycms/blob/master/core/lib/refinery/cli.rb#L8-L12

2015-02-18 15:05 GMT+01:00 Prokop Simek <notifications@github.com>:

prokopsimek:
```
2.1.5 :015 >   Refinery::Plugins.registered.pathnames.map{|p| %w(app vendor).map{|dir| puts p.join(dir, "views" )} }.flatten.uniq
/home/user/.rvm/gems/ruby-2.1.5/bundler/gems/refinerycms-i18n-d085574f6a29/app/views
/home/user/.rvm/gems/ruby-2.1.5/bundler/gems/refinerycms-i18n-d085574f6a29/vendor/views
/home/user/.rvm/gems/ruby-2.1.5/bundler/gems/refinerycms-6a3fb0afedf7/core/app/views
/home/user/.rvm/gems/ruby-2.1.5/bundler/gems/refinerycms-6a3fb0afedf7/core/vendor/views
/home/user/.rvm/gems/ruby-2.1.5/gems/refinerycms-wymeditor-1.0.6/app/views
/home/user/.rvm/gems/ruby-2.1.5/gems/refinerycms-wymeditor-1.0.6/vendor/views
[...]

PR #2900, opened 2015-02-18
Fix overriding hierarchy

prokopsimek:
Because of: https://github.com/refinery/refinerycms/issues/2899


simi:
Not a good idea at all. But we can change plugin registering to prepend new ones.

prokopsimek:
@parndt ?

parndt:
We should instead change the plugin registration code to `.unshift` rather than `<<` when registering a new plugin.

parndt:
Closed by #2992

PR #2902, opened 2015-02-18
fix strong params generation in engine generator

### LINKED ISSUE #2901: Engine generator broken for image and resource field types
Using refinery 2.1.5 with rails 4.1.8 

I created a new engine - using the engine generator - that had a photo field and noticed that the image wasn't being saved when creating/updating a record.
Command used:
`rails generate refinery:engine doctor name:string title:string photo:image youtube:string about:text`

This resulted in having this in the params white list:

    def doctor_params
          params.require(:doctor).permit(:name, :title, :photo, :youtube, :about)
    end

Noticed `Unpermitted parameters: photo_id` in the development log so I changed **photo** to **photo_id** in the params whitelist which solved the problem.

A resource field will experience the same behavior.

simi:
fixes #2901

### DIFF
--- a/core/lib/generators/refinery/engine/engine_generator.rb
+++ b/core/lib/generators/refinery/engine/engine_generator.rb
@@ -1,5 +1,6 @@
 require 'rails/generators/named_base'
 require 'refinery/extension_generation'
+require 'refinery/generators/generated_attribute'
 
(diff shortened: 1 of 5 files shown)


parndt:
Wonderful!

PR #2903, opened 2015-02-19
Extraction of refinerycms-authentication-devise

parndt:
Once this is released, it'll be reflected as a default gem in the Gemfile instead.

This should reduce our maintenance headache considerably because if someone wants to use their own Devise, they don't have to make Refinery no-op first.

Version 3.1.0: rewrite to have proper adapters in the core system that make switching to other systems trivial.  I wanted to do this for version 3.0.0 but then it'll *never* get released.


bricesanchez:
Wow, can't wait for that :smiley:

parndt:
Closed in favour of #2908

PR #2906, opened 2015-02-19
Move Globalize and Friendly_id dependencies in core 

bricesanchez:
If we require will_paginate and decorators in core, i think we should require globalize and friendly_id in core too, it will be better for us to update engines.

We will have to remove those dependencies in our engines.

### DIFF
--- a/authentication/refinerycms-authentication.gemspec
+++ b/authentication/refinerycms-authentication.gemspec
@@ -23,7 +23,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'refinerycms-core',  version
   s.add_dependency 'actionmailer',      rails_version
   s.add_dependency 'devise',            ['~> 3.0', '>= 3.2.4']
-  s.add_dependency 'friendly_id',       '>= 5.0.0.rc1'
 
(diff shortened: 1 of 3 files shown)


parndt:
Thanks for this.. unfortunately core doesn't use friendly_id, nor does it use globalize.

It's conceiveable that you'd want to use Refinery without pages, as you can actually just construct your own package like this:
```ruby
gem "refinerycms-images"
gem "refinerycms-resources"
```

^ that's a perfectly valid use of Refinery, in which case you won't need friendly_id or globalize in your object space

 ❤️💛💙💜💚💜💙💛❤️💛💙💜💚💜💙💛❤️❤️💛💙💜💚💜💙💛❤️💛💙💜💚💜💙💛❤️❤️💛💙💜💚💜💙💛❤️💛💙💜💚💜💙💛❤️❤️💛💙💜💚💜💙💛❤️💛💙💜💚💜💙💛❤️❤️💛💙💜💚💜💙💛❤️💛💙💜💚💜💙💛❤️❤️💛💙💜💚💜💙💛❤️💛💙💜💚💜💙💛❤️

bricesanchez:
Ok :sweat_smile:

PR #2907, opened 2015-02-21
Add scope for FriendlyId.

raskhadafi:


simi:
Hello @raskhadafi. Thanks for the patch. I wonder why tests are not failing without this. Can you provide some failing test (and passing with your patch)? If you don't have a time, just feel free to provide a minimal steps needed to reproduce and I can try to write it.

parndt:
This patch doesn't make sense to me; we already have a Finder class that does this under the hood and the by_slug method is a scope we provide.

https://github.com/refinery/refinerycms/blob/master/pages/lib/refinery/pages/finder.rb#L157

parndt:
Okay this makes some more sense to me after I've discovered that `by_slug` doesn't apply on top of an existing scope at all, it just finds by slug in the global scope..

PR #2913, opened 2015-03-03
Add guide to use StaffMember.friendly.find

ratanachai:
Without using friendly.find in Refinery 3.0 ActiveRecord::RecordNotFound is throw.


parndt:
Thanks, I've also applied this to the master branch.

Issue #2917, opened 2015-03-08
Page part title propagates into other page parts


anitagraham:
When a new page part (content section) is created a new editor tab is created with the tab header having the title of the page part.    If, before saving the page, the user looks at another page part, it will have the new page part name inserted into any lists it holds. Once the page is saved and refreshed the extra text is not present.

![new page part](https://cloud.githubusercontent.com/assets/397118/6545029/5751e01c-c5a9-11e4-95c4-0ab35ec26980.png)

PR to follow.

bricesanchez:
#2918 should fix this bug, thanks @anitagraham :)

PR #2918, opened 2015-03-08
Limit propagation of new part name/link for  #2917

anitagraham:
Between creation of new page part and save of page with new part, the
other editor tabs would have the link to the new page part added to
their content area.

This fix ensures the link is only added to the page tabs.

### DIFF
--- a/core/app/assets/javascripts/refinery/admin.js.erb
+++ b/core/app/assets/javascripts/refinery/admin.js.erb
@@ -609,7 +609,7 @@ var page_options = {
               // Add a new tab for the new content section.
               $('#page_part_editors').append(data);
 
-              page_options.tabs.find("ul").append("<li><a href='#page_part_new_"+$('#new_page_part_index').val()+"'>"+part_title+"</a></li>");
+              $("ul#page_parts").append("<li><a href='#page_part_new_"+$('#new_page_part_index').val()+"'>"+part_title+"</a></li>");
               page_options.tabs.tabs("refresh");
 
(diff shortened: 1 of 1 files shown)


parndt:
Hmm, this seems like it's supposed to be passed in. Can we limit the scope of the function to just apply to the one it's supposed to? I don't want to break this for other uses accidentally

anitagraham:
Do you mean like this:
````
page_options.tabs.find("ul#page_parts").append("<li><a href='#page_part_new_"+$('#new_page_part_index').val()+"'>"+part_title+"</a></li>");
````

parndt:
Yup, exactly.

anitagraham:
OK... but as I'm targeting an id the effect should be identical.

parndt:
Until you have two of these on the same page :smile:

anitagraham:
1. Aren't there are always just one set of page tabs on a page?
2. If it is an id there should only be one on the page. (Two identical ids will work most of the time, but not always)
I am changing it, btw, just arguing for the sake of it.

parndt:
Yeah, probably it should be a class

anitagraham:
I'd like to see an example of a container/page with two sets of tabs for content sections.

anitagraham:
Travis fail....whyowhyowhy.
<aside>
If you look at the generated HTML for refinery there are (lots, some, a few) occurrences of tags which have identical class and id attributes.  If one were scrubbing Refinery down then these would be candidates for a good cleanup. And then the CSS could be cleaned up too....
</aside>

PR #2919, opened 2015-03-08
Handle the response from delete_page_part

anitagraham:
The response from delete_page_part wasn't being handled. Added a handler which will update the flash notice with success or error message.


parndt:
Nice work!

PR #2921, opened 2015-03-10
Update 2 - Amazon S3 for Uploads.textile

ollaollu:
Added a direct override for the for the S3 configuration in the core.rb file if the ones in application.rb and environments/production.rb don't solve the problem as was the situation with me.


parndt:
@ollaollu this has inspired me to make this easier. Can you look at #2923

PR #2922, opened 2015-03-10
Fixing the Forms in Master

CJBrew:
I'm working on integrating the work I did for 2.1.0 on the Forms generator into Master.

This is 95% there, but I'm having some problems with the settings for the forms.

You'll notice 3(?) specs fail -- these are the ones that test setting Notification Recipients and the Confirmation message. The views for these use different approaches to set the parameters (not to mention different methods to create the form objects (text_area_tag rather than f.text_area etc). 

More info: notification recipients form gives params like
```
{ .., ... ,.., :settings => { value: "email_addresses" }, ... }
```

The message field form gives params like
```
{ ..., ..., :confirmation_message_subject => "blah", :confirmation_message_body => "blee" }
```
(those field names are probably not exactly correct. Point is, they're not in an anonymous "settings" hash).

I don't completely understand how the settings controller is supposed to handle these values particularly with regard to the Strong Parameters, or what the best design for the forms would be. I think this should probably follow the pattern of other Settings classes in the RefineryCMS system.

A quick review of this work and what needs to be done before you can accept this PR would be very welcome!


core/lib/generators/refinery/form/templates/app/controllers/refinery/namespace/plural_name_controller.rb.erb
```diff
@@ -46,10 +45,18 @@ module Refinery
...
-
+    private
+
+    # Only allow a trusted parameter "white list" through.
+    def <%= "#{singular_table_name}_params" %>
+    <%- if attributes_names.empty? -%>
+            params[:<%= singular_table_name %>]
+          <%- else -%>
+            params.require(:<%= singular_table_name %>).permit(<%= attributes.map { |attr| ":#{attr.column_name}" }.join(', ') %>)
+          <%- end -%>
```
parndt: the indentation seems a bit funky here

core/lib/generators/refinery/form/templates/app/models/refinery/namespace/setting.rb.erb
```diff
@@ -1,10 +1,10 @@
 module Refinery
   module <%= namespacing %>
-    class Setting < Refinery::Core::BaseModel
+    class Setting
```
parndt: good call!

core/lib/generators/refinery/form/templates/spec/controllers/refinery/plural_name/plural_name_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        it "redirects to thank_you" do
+          post :create
+          response.should redirect_to "/<%= plural_name %>/thank_you"
+        end
+
+        describe "when it can't save the <%= singular_name %>" do
+
+          before(:each) {
+          allow_any_instance_of(<%= class_name %>).to receive(:save).and_return( false )
```
parndt: indentation, and `before(:each)` could just be `before`

core/lib/generators/refinery/form/templates/spec/features/refinery/namespace/singular_name_spec.rb.erb
```diff
@@ -0,0 +1,35 @@
...
+            <% if a.type == :string -%>
+            fill_in "<%= singular_name %>_<%= a.name %>", :with => "Test"
+                 <% elsif a.type == :text -%>
+            fill_in "<%= singular_name %>_<%= a.name %>", :with => "Test"
+                 <% end -%>
+               <% end -%>
+             click_button ::I18n.t('.refinery.<%= plural_name %>.<%= plural_name %>.new.send')
+
+               page.should have_content("Thank You")
+               Refinery::<%= namespacing %>::<%= class_name %>.count.should == 1
```
parndt: indentation in the preceeding 10 lines.

parndt:
This seems pretty great, I've just left some comments :smile:

CJBrew:
@parndt -- now passing on Travis as I've added support (minimal) for types like "Select", "Radio", "Dropdown" etc. 

I think I've fixed the indentation issues you mentioned, but there are probably some more that appear in the generated files and I find tricky to get right with a sensible balance between readable generator code. Do you have an approach or tool for that other than inspection and just removing spaces and extra line breaks?

parndt:
Nah, usually I just run the generators and inspect he indentation manually :smile:
Thanks for tackling all of this work! I'll review this in the next few days.
Hopefully, other @refinery/collaborators can too.

CJBrew:
I ran RuboCop against the generated code. It came up with tons of stuff,
probably too many changes unless you care about things matching their style
guidelines.

Having spent far too much time in the past making FXCop happy about style,
I think it's probably only worthwhile using a 'linter' to catch potential
coding errors.

cheers,
Chris

CJBrew:
Intended to fix issues in https://github.com/refinery/refinerycms/issues/2690

parndt:
@CJBrew I've opened a new PR which is rebased against master with some of the newer changes in #3009

PR #2923, opened 2015-03-10
S3 doc update

parndt:
This comes off the back of #2921 but this makes it easier for people to get it working.

@ollaollu would this work for you?


ollaollu:
@parndt Oh great. Yes it does work for me.

PR #2924, opened 2015-03-11
Switched site bar background gradient for a solid colour

parndt:
Before:
screen shot 2015-03-11 at 17 33 50

After:

screen shot 2015-03-11 at 17 33 23


anitagraham:
:+1: Nice dark blue. #414550
Is there (or will there be) a list of colours used in Refinery? (I used _colours.scss for the icon colours.)

parndt:
@anitagraham that'd be good.

PR #2925, opened 2015-03-13
Resources: Use permitted params, allow method to be overridden

ngonzalez:


parndt:
@ngonzalez can you please change the implementation of resource_params to this:

```ruby
def resource_params
  # update only supports a single file, create supports many.
  if action_name == 'update'
    params.require(:resource).permit(:file)
  else
    params.require(:resource).permit(:file => [])
  end
end
```

Then your PR will be ready to merge.

ngonzalez:
@parndt Thanks for the comment, I did the changes you mentioned:
https://github.com/refinery/refinerycms/commit/2c88590acbbe8ef625bec24d7aae89b159e6d2fc

PR #2935, opened 2015-03-21
Restyled big orange site bar buttons to remove orange.

parndt:
Seems to be where things are heading

Before:
screen shot 2015-03-21 at 16 51 39

After:
screen shot 2015-03-21 at 16 50 11

### DIFF
--- a/core/app/assets/stylesheets/refinery/site_bar.css.scss
+++ b/core/app/assets/stylesheets/refinery/site_bar.css.scss
@@ -27,30 +27,14 @@
     #editor_switch {
       text-align: center;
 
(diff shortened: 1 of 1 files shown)


simi:
:+1:

parndt:
We should also consider a different font

anitagraham:
Font: for the whole system or just the site bar buttons?
To do?
better mobile readability?
distinguish active item from descriptive text?

parndt:
Whole system, dunno, just an offhand comment

anitagraham:
aaah... for fun!

PR #2941, opened 2015-04-05
Google Analytics only if Rails.env.production?

prokopsimek:
What about this one? Most of time we don't want to track on other environment than production...

### DIFF
--- a/core/app/views/refinery/_head.html.erb
+++ b/core/app/views/refinery/_head.html.erb
@@ -9,6 +9,6 @@
 <%= stylesheet_link_tag "application" %>
 <%= yield :stylesheets %>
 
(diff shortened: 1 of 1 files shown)


parndt:
It currently only works if the page code is set and if it's not a local_request? (which it usually is in development mode) and you're not logged in.  This code is in the partial itself.

Is this not working?

prokopsimek:
@parndt It is, but if you want to use more than only development & production environment... E.g. "staging" or "beta"?

parndt:
In this case perhaps it would be better to set the analytics code via ENV and neglect to set it in the staging environment?

prokopsimek:
ok

PR #2942, opened 2015-04-08
Pages form fields after...

bricesanchez:
When i add more custom menus, i'd to override the all _form_advanced_options instead of just adding a partial like form_fields_after_title.

BTW, I've change the way we render those partials, no more empty partial :)

### DIFF
--- a/pages/app/views/refinery/admin/pages/_form.html.erb
+++ b/pages/app/views/refinery/admin/pages/_form.html.erb
@@ -10,7 +10,9 @@
     <%= f.text_field :title, :class => "larger widest" %>
   </div>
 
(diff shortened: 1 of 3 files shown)


parndt:
Personally I'd rather we just rendered empty partials. :confused:

bricesanchez:
Is the best should be moving all menus in a partial ? 
Or use form_fields_after_show_in_menu empty partial?

For me, it's strange to create empty partials but i can live with that :D

parndt:
Maybe the partial needs a more expressive name of what it is for, for example `form_fields_for_checkbox_options` or something like that

PR #2928, opened 2015-03-14
Apply fix to controller macros suggested by parndt

CJBrew:
- to remove deprecation warnings in rspec run

Sending pull request in order to testing this in the Travis CI build -- 

it may fix https://github.com/refinery/refinerycms/issues/2927 -- my tests seem to pass.

PR #2946, opened 2015-04-09
Use engine routes in specs.

### LINKED ISSUE #2937: Running pages spec after core/spec/controllers/refinery/admin/dummy_controller_spec.rb fails
To reproduce, locally:

```
bundle exec rspec core/spec/controllers/refinery/admin/dummy_controller_spec.rb pages/spec --order standard
```

I'm guessing this has something to do with #2928

simi:
fixes #2937

ping @parndt, @CJBrew

```
[ryba@retro  refinerycms (use-engine-routes-in-specs)]❤ bin/rspec core/spec/controllers/refinery/admin/dummy_controller_spec.rb pages/spec --order standard
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}
..........................................................................DEPRECATION WARNING: You are calling a `*_path` helper with the `only_path` option explicitly set to `true`. This option will stop working on path helpers in Rails 5. Simply remove the `only_path: true` argument from your call as it is redundant when applied to a path helper. (called from block (3 levels) in <module:Refinery> at /opt/ryba/refinerycms/pages/spec/features/refinery/pages_spec.rb:231)
...................................................................................................................................................................

Finished in 1 minute 26.79 seconds (files took 0.73418 seconds to load)
237 examples, 0 failures
```


CJBrew:
Hi Josef,

Thanks for the fix. Not saying it's wrong, but I don't understand it 100%.
Especially the bit where you build the routes

Not by a computer for a few days, but could you briefly explain this for my
benefit?

Does this change basically achieve the same thing I did in nicer, up to
date code style with reference to the Refinery Engine? Is this what I
missed  when I did it (using the examples on the Rspec documentation as
inspiration)?

Cheers, Chris

simi:
Sorry, but I'm out of time. Check https://github.com/refinery/refinerycms/pull/2928/files#r28077110, that declaration belongs to engine code and you're changing testing code only. Actually it's future goal to have isolated "Frontend" and "Backend" namespaces.

Issue #2685, opened 2014-09-09
Engine generator image type not creating image add link


AlterEgo7:
Following the getting started guide and upon running the following generator command:

```
rails generate refinery:engine event title:string date:datetime photo:image blurb:text
```

when I navigate to the add event page in the photo field there is a text field instead of a link for adding images.

![snapshot1](https://cloud.githubusercontent.com/assets/4788639/4204986/16fdc4d6-383c-11e4-9c81-e50af171fa67.png)

bluefishers:
I meet the same issue.
Environment:
Ubuntu 12.04
Rails 4.1.8
ruby 2.0.0

razvanh:
I have the same issue

My Gemfile:

```
source 'https://rubygems.org'
gem 'rails', '4.1.6'
group :development, :test do
  gem 'sqlite3'
end
gem 'sass-rails', '~> 4.0.3'
gem 'uglifier', '>= 1.3.0'
gem 'coffee-rails', '~> 4.0.0'
gem 'jquery-rails'
gem 'turbolinks'
gem 'jbuilder', '~> 2.0'
gem 'sdoc', '~> 0.4.0',          group: :doc
gem 'spring',        group: :development
gem 'refinerycms', git: 'https://github.com/refinery/refinerycms', branch: 'master'
gem 'quiet_assets', :group => :development
gem 'refinerycms-acts-as-indexed', ['~> 2.0', '>= 2.0.0']
gem 'refinerycms-wymeditor', ['~> 1.0', '>= 1.0.6']
gem 'bootstrap-sass', '~> 3.3.4.1'
gem 'refinerycms-doctors', path: 'vendor/extensions'
```

Ruby 2.1.5
bash
Ubuntu 14.04.1 LTS

Using cloud9.io

parndt:
@razvanh can't reproduce using exactly that example at cloud9: https://ide.c9.io/parndt/refinery-image-issue

```erb
  <div class='field'>
    <%= f.label :photo -%>
    <%= render '/refinery/admin/image_picker',
               :f => f,
               :field => :photo_id,
               :image => @event.photo,
               :toggle_image_display => false -%>
  </div>
```

developpando:
Hi this is happening to me as well. Has anyone found a solution yet?
I'm on Debian GNU/Linux 7.8 (wheezy) 
Thank you
Ang

parndt:
I'm finding it really hard to reproduce the issue :(

[...]

PR #2948, opened 2015-04-11
Fixes issue #2685

parndt:
In some instances, files are loaded in a different order and, given the database migrations were modifying the properties of the attributes, it was setting them back to `integer` and `photo_id` instead of `image` and `photo`, so an image picker would not be displayed where it should be.


simi:
I have an idea how to solve this in #2951 also.

parndt:
REPLACED by #2951

PR #2949, opened 2015-04-11
Added rspec-retry to avoid CI failures for timing issues.

parndt:
@simi what do you think?


simi:
It is really sad (and attached cat is also) we need this, but it will be better for us to use our time to do better things than restarting CI jobs.

59d5403649f73e874b8a81c197faa71c-300x268

PR #2953, opened 2015-04-16
Update guide Menu Presenter for Refinery 3.0

bricesanchez:


doc/guides/9 - Presenters/1 - Menu Presenter.textile
```diff
@@ -41,31 +42,48 @@ Run the migration:
...
+Refinery::Admin::PagesController.class_eval do
+  # work around from https://github.com/refinery/refinerycms-page-images/blob/master/app/decorators/controllers/refinery/admin/pages_controller_decorator.rb
+  if self.instance_methods.exclude?(:page_params_with_show_in_footer_params)
+    def page_params_with_show_in_footer_params
+      show_in_footer_params = params.require(:page).permit(:show_in_footer)
+    end
+
+    alias_method_chain :page_params, :show_in_footer_params
+  end
+end
```
parndt: This workaround shouldn't be required since Decorators 2.0.0 was released, and you're ignoring all of the params that were previously allowed.  Try this:

```ruby
def page_params_with_show_in_footer
  page_params_without_show_in_footer.permit(:show_in_footer)
and
alias_method :page_params, :show_in_footer
```

bricesanchez:
Fixed, thanks for the tip, i didn't know that the workaround was obsolete.

bricesanchez:
As i understood it, your tip doesn't work : 

```ruby
Refinery::Admin::PagesController.class_eval do
  def page_params_with_show_in_secondary_menu
    page_params_without_show_in_secondary_menu.permit(:show_in_secondary_menu)
  end
  alias_method :page_params, :show_in_secondary_menu

  def page_params_with_show_in_main_menu
    page_params_without_show_in_secondary_menu.permit(:show_in_main_menu)
  end
  alias_method :page_params, :show_in_main_menu

  def page_params_with_show_in_footer_menu
    page_params_without_show_in_secondary_menu.permit(:show_in_footer_menu)
  end
  alias_method :page_params, :show_in_footer_menu
end
```

doc/guides/9 - Presenters/1 - Menu Presenter.textile
```diff
@@ -41,31 +42,44 @@ Run the migration:
...
+h3. Allow +show_in_footer+ attribute in +Refinery::Admin::PagesController+
+
+Before overriding Refinery's form view, we want to decorate the +Refinery::Admin::PagesController+ class. Create a file +app/decorators/controllers/refinery/admin/pages_controller_decorator.rb+ with this content:
+
+<ruby>
+Refinery::Admin::PagesController.class_eval do
+  def page_params_with_show_in_footer
+    page_params_without_show_in_footer.permit(:show_in_footer)
+  and
+  alias_method :page_params, :show_in_footer
```
simi: Is this meant to be `alias_method_chain`?

doc/guides/9 - Presenters/1 - Menu Presenter.textile
simi: You mean `alias_method_chain`?

bricesanchez:
if i use alias_method_chain it cause a loop error :

SystemStackError (stack level too deep):
  app/decorators/controllers/refinery/admin/pages_controller_decorator.rb:3:in `page_params_with_show_in_secondary_menu'
  app/decorators/controllers/refinery/admin/pages_controller_decorator.rb:8:in `page_params_with_show_in_main_menu'
  app/decorators/controllers/refinery/admin/pages_controller_decorator.rb:13:in `page_params_with_show_in_footer_menu'
  app/decorators/controllers/refinery/admin/pages_controller_decorator.rb:3:in `page_params_with_show_in_secondary_menu'
  app/decorators/controllers/refinery/admin/pages_controller_decorator.rb:8:in `page_params_with_show_in_main_menu'
  app/decorators/controllers/refinery/admin/pages_controller_decorator.rb:13:in `page_params_with_show_in_footer_menu'
  app/decorators/controllers/refinery/admin/pages_controller_decorator.rb:3:in `page_params_with_show_in_secondary_menu'

parndt:
@bricesanchez which version of decorators is your project using?

bricesanchez:
It uses decorator 2.0.1, i didn't check the syntax of my code today, i will have time tomorrow, i will let you know if i still have this error.

bricesanchez:
As @anitagraham said on gitter, this snippet works for her and me : 
```ruby 
Refinery::Admin::PagesController.class_eval do
  def page_params_with_menus_params
    menus_params = params.require(:page).permit(:show_in_secondary_menu,
                                                :show_in_main_menu,
                                                :show_in_footer_menu)
  end

  alias_method_chain :page_params, :menus_params
end
```

parndt:
Yeah but doesn't that mean all other attributes will be ignored?

bricesanchez:
I don't know how to check if they are ignored but it looks like to work well.

parndt:
Well they will be ignored, because this method you've proposed overrides the original method without calling the original method.

bricesanchez:
So,what is the good way? I was using the workaround until your advice. I don't know what to think here.

anitagraham:
I use a merge:
````ruby
testimonial_params = params.require(:page).permit(:testimonials_show, :testimonials_count, :testimonials_select)
page_params_without_testimonial_params.merge(testimonial_params)
````

**Comment by parndt** [comment]
Ah, you can merge those? I thought you had to call permit again.

**Comment by parndt** [comment]
Okay, I've figured out the basic trouble and I will add a patch to this branch.

**Comment by parndt** [comment]
@simi @bricesanchez @anitagraham I think it's much nicer now?

**Comment by anitagraham** [comment]
I'm having trouble getting this to work. Will it work with the application and an extension using this technique?

**Comment by parndt** [comment]
It should do, I just tried it and it worked for me. (I know, worst ever response).  Let me know what trouble you're having?

**Comment by anitagraham** [comment]
I can set and unset the page.show_in_footer successfully on its own.

I've added the same pattern to refinerycms-testimonials thus:
````ruby
module RefineryAdminPageParams
  def permitted_page_params
    super << [:testimonials_show, :testimonials_count, :testimonials_select]
  end
end
Refinery::Admin::PagesController.prepend RefineryAdminPageParams
````

If I add refinerycms-testimonials to the gemfile (bundle gemfile etc) I can no longer set the page.show_in_footer, and see
`Unpermitted parameter: show_in_footer`  in the log, and the parameter is not set.

Here: (https://gist.github.com/anitagraham/ed3b73e450cffed70fc6#file-show-in-footer-not-working-L21)
and here: (https://gist.github.com/anitagraham/ed3b73e450cffed70fc6#file-show-in-footer-working-L39)

**Comment by parndt** [comment]
Is that because you're using the same module name?

**Comment by anitagraham** [comment]
Probably... but then the example should name its module like this
````ruby
module RefineryAdminPageShowInFooterParams
````
(Params optional perhaps)

**Comment by parndt** [comment]
@anitagraham I was thinking maybe we should do this instead:

```ruby
Refinery::Admin::PagesController.prepend(
  Module.new do
    def permitted_page_params
      super << :show_in_footer
    end
  end
)

```

anitagraham:
Module.new  what happens to Module.old? 
(I could look it up but then  it makes the discussion less enlightening to those reading it a year later...)

parndt:
Haha.. it just creates a new anonymous module with this method and prepends it onto the class, thus making it so that `super` still works.

anitagraham:
Hokay.  Also did you notice - as I just did looking for Module.new - that the Google Doodle today is Dame Ngaio Marsh? A New Zealand Google doodle.

parndt:
Nice!

@anitagraham does this work with your example from before?

anitagraham:
Yes!! :+1:

PR #2954, opened 2015-04-20
Make FormHelper extension required_label behave the same as label for i18n

johanb:
# For your consideration: #

Refinery has a `required_label` form helper method. This intended behavior of `required_label` seems to be to append a " *" to the content and add a required class. Currently however it changes Rails default fallback behavior of the `label` method for forms.

From the [docs](http://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-label):
> Returns a label tag tailored for labeling an input field for a specified attribute (identified by method) on an object assigned to the template (identified by object). **The text of label will default to the attribute name unless a translation is found in the current I18n locale (through helpers.label.<modelname>.<attribute>) or you specify it explicitly**. Additional options on the label tag can be passed as a hash with options. These options will be tagged onto the HTML as an HTML element attribute as in the example shown, except for the :value option, which is designed to target labels for #radio_button tags (where the value is used in the ID of the input tag).

_emphasis mine_

Refinery's current implementation does not respect this behavior and jumps straight to the AR method described in the rails docs:

I've made a PR which changes this behavior to mimic the Rails fallback behavior without changing the arity of the method definition of `required_label` (for compatibility sake).

While changing this behavior I couldn't find existing specs, is this correct or did I look in the wrong places ?

## TODO ##
* [ ] Spec this
* [ ] Discuss possibility of changing the arity to be compatible with the `label` method

# Update #
I'd actually prefer to remove this entire thing from Refinery, it solves something so trivial. I'd be happy to change this PR to that.. Seems it's not used in this: https://github.com/refinery/refinerycms/pull/2922


parndt:
@johanb removing it sounds great.

johanb:
@parndt I've removed it.

This is however used here: https://github.com/refinery/refinerycms-inquiries/blob/master/app/views/refinery/inquiries/inquiries/_form.html.erb you want me to whip up a PR for the extensions hosted under "refinery" that depend on this to ?

parndt:
@johanb that'd be lovely, thanks!

johanb:
Accidentally removed my branch, which closed the PR. Sorry

parndt:
Merged into a single commit b4b20cd, thanks!

ghoppe:
This breaks `rails g refinery:form`

johanb:
Missed that @ghoppe, will investigate. Thanks for reporting it.

PR #2959, opened 2015-04-24
Iconfont fixes

anitagraham:
1. Move Rails.application.config.assets.precompile settings to core/config/assets.rb
2. Specify css asset targets to avoid compiling partials
3. Fix padding for flash message
4. Stop lowercasing flash close button text


anitagraham:
Ran tests locally and seemed to trigger  errors in TranslatePages tests

core/config/initializers/assets.rb
```diff
@@ -1,6 +1,27 @@
+# These precompile sets are split for readablity
+# scripts
+ Rails.application.config.assets.precompile += %w(
```
simi: Why we can't leave this hidden in engine.rb?

anitagraham: This is now the place where someone following the Rails recommendations would expect to find them.
I'm not sure why "hidden in engine.rb" is a good idea.

simi: This is expected to be in `initializer` in Rails engine IMHO.

see https://github.com/activeadmin/activeadmin/blob/c6347ea2429d0f052e167ba472a35f1d1061124e/lib/active_admin/engine.rb#L3-L10 for example what I mean

@parndt what do you think?

parndt: Hidden in engine gets my vote

anitagraham: As you wish, but (whine) whyyyy?

parndt:
Looks good, just needs rebasing which I'll do later

simi:
@parndt is rebase still needed? I have active "free iPad" button here.

parndt:
Yes, a single commit only

core/lib/refinery/core/engine.rb
```diff
@@ -79,15 +79,39 @@ def refinery_inclusion!
...
-          refinery/*
+          refinery/*.js
           refinery/icons/*
           modernizr-min.js
           admin.js
         )
+
+        # stylesheets
+        Rails.application.config.assets.precompile += %w(
+          refinery/refinery.css
```
bricesanchez: Why don't we allow *.css as we do on *.js ?

At this time, i've to add `refinery/*` in my `Rails.application.config.assets.precompile` in assets.rb to be able to use refinery extension who have CSS/JS assets.

anitagraham: @bricesanchez I didn't see this line note till today.
We don't allow *.css because we are using SASS style stylesheets and partials.
The stylesheet (in this case `refinery.scss`) uses `@import` to import the partials which make up the complete stylesheet. 
If you use `refinery/*.css` the precompiler will try to compile the partials on their own, leading to errors.

An extension should add its own assets to the precompile array, and you will automatically acquire them.

bricesanchez:
@anitagraham Thanks for your feedback, my last comment is wrong. Your solution and your work on css works well :)

PR #2960, opened 2015-04-24
Change the text of the search field

anitagraham:
1. Change long, awkward string to short string suitable as placeholder
2. Add longer string to use as tooltip

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -56,7 +56,7 @@ en:
         no_results: 'Sorry, no results found'
         cancel_search: 'Cancel search'
         search_input_notice: "Search&#8230"
-        search_input_help: 'Enter a word or name you want to search for'
+        search_input_help: 'Enter a phrase to search for'
       delete:
         message: "Are you sure you want to remove '%{title}'?"
       error_messages:
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -23,7 +23,7 @@ en:
     admin:
       menu:
         reorder_menu: Reorder menu
-        reorder_menu_done: I'm done reordering the menu
+        reorder_menu_done: I've finished reordering the menu
       dialogs:
         show:
           save: Save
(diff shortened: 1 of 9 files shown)


core/config/locales/en.yml
```diff
@@ -51,11 +51,12 @@ en:
...
         current: Current File
       search:
-        button_text: Search
+        button_text: Go
         results_for_html: "Search results for &#8216;<em>%{query}</em>&#8217;"
         no_results: 'Sorry, no results found'
         cancel_search: 'Cancel search'
-        search_input_notice: "Put here a string of at least 3 characters containing what you are looking for."
+        search_input_notice: "Search&#8230"
+        search_input_help: 'Enter a word or name you want to search for'
```
simi: s/word or name/phrase/g

core/config/locales/en.yml
```diff
@@ -51,11 +51,12 @@ en:
         name: Add File
         current: Current File
       search:
-        button_text: Search
+        button_text: Go
```
parndt: I really think `Search` makes more sense.

anitagraham: It does - on it's own. However with "Search" you end up with this:
screen shot 2015-04-27 at 10 19 32 am

which also looks odd. Happy for a better term than "Go" for sure. 
Or just have the search triggered by Enter and no button.

parndt: Maybe the placeholder should change

bricesanchez: According this website : http://www.w3schools.com/tags/att_input_placeholder.asp
"The placeholder attribute specifies a short hint that describes the expected value of an input field (e.g. a sample value or a short description of the expected format)."

We should write "Text", "Keywords", ... instead of "Search..."

parndt: I agree with that

parndt: Preferably a related phrase that we already have translated!

anitagraham: How about "Search for..." or "Search terms"  (expected audience experienced enough to understand 'search terms').

Translations: Given that we are not changing the meaning of the field, just providing a more fluent and less verbose English translation, what do we need to do wrt other languages.  I had a look at `it.yml` and they do seem to have translated word for word.
Do we alert them to changed fields? (which them?) 

Also, I would anticipate that in the nearish future, I would like to re-present this field, with a placeholder in the field and a search icon on the button. (and the parts connected smoothly to each other).

I have to go now, but will perhaps start a conversation about backend presentation in a better spot than this.

anitagraham:
See PR https://github.com/refinery/refinerycms-acts-as-indexed/pull/5 which uses these search strings.

parndt:
Looks like we need to change `search_input_notice` to `search_input_help` on all translation files still

Issue #2715, opened 2014-10-20
Generated migrations shouldn't reference the Refinery object 


Aerlinger:
There are two migrations installed by Refinery that make calls to the base Refinery object: the `AddCustomSlugToRefineryPages` migration and the `CreateRefinerycmsPagesSchema` migration. IMHO this is bad form as it introduces unnecessary coupling between migrations and the Refinery gem. For instance, if one were to later remove Refinery from the Gemfile, it would not be possible to run Refinery's migrations in a new environment as the Refinery object would no longer be defined. To fix this,  one would have to retroactively remove/update these old migrations which can be dangerous in a production environment.

If possible, these migrations should be refactored to not rely on making direct calls to Refinery.

PR #2965, opened 2015-04-30
Add ability to translate image_title and image_alt fields

bricesanchez:


images/app/helpers/refinery/admin/images_helper.rb
```diff
@@ -19,6 +19,12 @@ def thumbnail_urls(image)
 
         thumbnail_urls
       end
+
+      # We show the title from the next available locale
+      # if there is no title for the current locale
```
simi: I'm not sure if this is what users wants everytime.

parndt: Agree, only when a fallback is configured should we fallback, and globalize should do that.

bricesanchez: So, this method is not good too : https://github.com/refinery/refinerycms/blob/master/pages/app/helpers/refinery/admin/pages_helper.rb#L50 ?

bricesanchez: Do you have any clue to fix this?

parndt: We do that on the backend because otherwise it is simply blank, but we don't want to display wrong content on the frontend.  Maybe we should write the pages one better too.

images/spec/support/shared contexts/admin_images_tab.rb
```diff
@@ -7,11 +7,11 @@
...
-  let(:gridview_title_selector) {'[tooltip]'}
-  let(:gridview_alt_selector)   {'[alt]'}
-  let(:listview_title_selector) {' > span.title'}
-  let(:listview_alt_selector)   {' > span.alt'}
+  let(:index_item_selector)        {'#records li'}
+  let(:gridview_img_selector)      {' > img'}
+  let(:gridview_title_selector)    {'[tooltip]'}
+  let(:gridview_alt_selector)      {'[alt]'}
+  let(:listview_title_selector)    {' > span.title'}
+  let(:listview_filename_selector) {' > span.preview'}
```
parndt: this changeset is a great example of why aligning things isn't ideal :smile:

images/app/helpers/refinery/admin/images_helper.rb
bricesanchez: At this time we only use `image_title_with_translations` method in backend.

images/spec/support/shared contexts/admin_images_tab.rb
bricesanchez: Yeah, good to know :)

bricesanchez:
@parndt Are you ok to merge this PR ?

parndt:
I note there are no tests specific to this feature?

bricesanchez:
Is someone could help me to test this PR ?

parndt:
@anitagraham do you have any ideas? I know you've been working on images specs lately.

anitagraham:
Yes, having a look at how the translations for page title are tested.
Testing failures due to incompatibility between Rspec and rspec-retry. (See gitter comment) make it a bit slower.

bricesanchez:
Thanks @anitagraham, it's really appreciated!

parndt:
So, to confirm: is the point of this feature so that the user can avoid uploading an image per locale?

bricesanchez:
Yes.

bricesanchez:
i will update this really soon PR :)

bricesanchez:
It's ready @parndt :)

images/db/migrate/20150430171341_translate_refinery_images.rb
```diff
@@ -0,0 +1,14 @@
...
+    Refinery::Image.create_translation_table!({
+      image_alt: :string,
+      image_title: :string
+    }, {
+      :migrate_data => true
+    })
+  end
+
+  def self.down
+    Refinery::Image.drop_translation_table! migrate_data: true
```
parndt: I hate to say it but this is going to suffer the same problems as #2715 - can you please apply the same solution as in 8f56bf47a7ecd6d348b26d9531ff5e4eb03d5f68 ?

parndt:
@bricesanchez sorry, I missed the migration problems before :frowning:

bricesanchez:
@parndt no problems, is it ok like that now?

parndt:
OK, I'm happy with this.  
@simi what about you?

PR #2966, opened 2015-04-30
Add ability to customize and translate resource name

bricesanchez:


resources/config/locales/en.yml
```diff
@@ -13,6 +13,8 @@ en:
           or: or
           replace: " replace it with this one..."
           maximum_file_size: The maximum file size is %{bytes}.
+          resource_title: Title
+          resource_title_help: Extra information about the file
```
anitagraham: Can I suggest that `resource_title_help` be "The title visible to users"?

Maybe?

anitagraham:
I guess resources don't require an alt attribute.

resources/config/locales/en.yml
bricesanchez: Yes it will be better !

bricesanchez:
No i don't think resources require an alt attribute. :)

bricesanchez:
output_ebnhpj

parndt:
I don't see any tests specific to this feature?

bricesanchez:
Is someone could help me to test this PR ?

anitagraham:
Will have a look @bricesanchez .  Refinerycms-pages has testing for translated page titles. That's where I am going to look.

anitagraham:
In some places you use `resource_title`, in others `title`.

bricesanchez:
Yes, i've tried to follow how it worked before. Do you have any advice in order to simplify the feature ?

anitagraham:
@bricesanchez Here is a test for the helper. It mostly comes from refinerycms-pages tests for translated page titles. 
````ruby
# Encoding: UTF-8
#/refinerycms/resources/spec/helpers/refinery/admin/resources_helper_spec.rb

require "spec_helper"

module Refinery
  module Admin
    describe ResourcesHelper do
      let(:resource) { FactoryGirl.build(:resource) }

      before do
        Globalize.with_locale(:en) do
          resource.resource_title = "draft"
          resource.save!
        end

        Globalize.with_locale(:lv) do
          resource.resource_title = "melnraksts"
          resource.save!
        end
      end

      describe "#resource_title_with_translations" do
        context 'when the resource has a title in the current locale' do
          it "returns the resource title" do
            expect(helper.resource_title_with_translations(resource)).to eq("draft")
          end
        end

        context 'when the resource does not have a title in the current locale' do
          it 'returns the title from the next locale with a title' do
            Resource.translation_class.where(:locale => :en).first.destroy
            expect(helper.resource_title_with_translations(resource)).to eq("melnraksts")
          end
        end
      end
    end
  end
end
````
This line `Resource.translation_class.where(:locale => :en).first.destroy` doesn't seem to work.
Today I don't have time to follow it up any further. 

I wonder if it would be better to just use the default title (file name) in locales without a translation.

bricesanchez:
@parndt Updated with tests :D
@anitagraham You're right! Now, i display the filename if there is no translation.

PR #2967, opened 2015-05-04
Fix resource-picker to use icons correctly

anitagraham:
Tests ran correctly on local system.


parndt:
Of course no tests failed when this partial was broken before 😅

anitagraham:
That is true and very, very sad. :weary:
So we need a test for resource-picker

PR #2972, opened 2015-05-06
alignment correction

mhtremblay:
Before:
screen shot 2015-05-06 at 12 54 24 pm

After:
screen shot 2015-05-06 at 12 54 01 pm


parndt:
Thanks!

Issue #2973, opened 2015-05-07
Can't remove all plugins from a user


johanb:
It currently won't let you remove the last plugin a user has access to. Most probably this is caused by strong parameters. I'll look into this and see if I can submit a PR.

PR #2974, opened 2015-05-07
Prevent crash if a user has no plugin's yet and you want to assign some.

johanb:
If a user is created in the database without any plugins, you can login etc but you don't have rights to any plugins. If later you try to add permissions to these plugins the application crashes because `max` yields `nil` instead of 0. 

This tiny change fixes this.

I can add a spec for this behavior if needed.

Probably also makes this code obsolete:
https://github.com/refinery/refinerycms/blob/master/authentication/app/models/refinery/user.rb#L59-L63

### DIFF
--- a/authentication/app/models/refinery/user.rb
+++ b/authentication/app/models/refinery/user.rb
@@ -56,14 +56,8 @@ module Refinery
       return unless persisted?
       plugin_names = string_plugin_names(plugin_names)
 
-      if plugins.empty?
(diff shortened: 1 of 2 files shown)


parndt:
This fixes #2973 ?

johanb:
@parndt it doesn't. do you want me to fix that in the same PR ?

parndt:
@johanb no, just checking

PR #2975, opened 2015-05-07
Allow a user to have 0 plugins

### LINKED ISSUE #2973: Can't remove all plugins from a user

johanb:
Rails is smart, when you unselect all checkboxes from a form it will post the checkboxes as `nil`. Strong parameters however expects you to send something of type `[]` instead of `nil` and blocks this parameter by filtering it out. When you update a active record model with out containing a key for `plugins` however it will not remove the existing plugins. This all works fine as long as you have at least 1 plugin, but when you want the user to have no plugins, this fails. Personally I think this is a strong_parameters bug. Have been fixing it in many projects.

This fixes #2973

@parndt would you like a feature spec to go along with this ?

### DIFF
--- a/authentication/app/controllers/refinery/admin/users_controller.rb
+++ b/authentication/app/controllers/refinery/admin/users_controller.rb
@@ -131,6 +131,7 @@ module Refinery
       end
 
       def user_params
+        params[:user][:plugins] ||= []
         params.require(:user).permit(
           :email, :password, :password_confirmation, :remember_me, :username,
           :login, :full_name, plugins: []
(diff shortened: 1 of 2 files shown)


parndt:
@johanb if you're willing to write one, yes please!

johanb:
Sure, will do

johanb:
@parndt done.

simi:
This is good to merge here IMHO. But is there similar bug already reported to strong params or rails? Maybe we can also open issue there.

PR #2977, opened 2015-05-09
Use remove on resource-picker, not delete

anitagraham:
The action ‘delete’ uses the delete method.
Action 'remove' here triggers javascript and leaves the unlinking of the resource until ‘Save’ happens.
Using an unlink icon instead of the standard delete. 
remove action using unlink icon

### DIFF
--- a/core/app/assets/stylesheets/refinery/_icons.scss
+++ b/core/app/assets/stylesheets/refinery/_icons.scss
@@ -41,6 +41,7 @@
 .info_icon            {@include icon('info-circle');}
 .page_icon            {@include icon('file-o',$icon_page_colour)}
 .preview_icon         {@include icon('eye', $icon_preview_colour)}
+.remove_icon          {@include icon('unlink', $icon_delete_colour);}
 .reorder_done_icon    {@include icon('thumbs-o-up')}
 .reorder_h_icon       {@include icon('exchange');}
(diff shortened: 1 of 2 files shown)


parndt:
One of these days we should add tests for this widget.. 

This is great, but clicking remove on an edit screen provides no opportunity to add another resource because the add button isn't visible.

anitagraham:
Was the add button visible before?
I can make it appear (and perhaps function :wink:) when the item is removed.

parndt:
@anitagraham it probably wasn't.. that'd be lovely!

anitagraham:
It works so well if I just leave out the 'if resource' on the code for the add button.
If there's no resource you can add one.
If there is a resource the add replaces the current resource.

parndt:
I guess as long as users are aware that adding a resource will replace the current resource that'd be fine.  I think it's reasonable when there is only a single resource attached to something that it'd be the case.

anitagraham:
Playing with this a bit more, I'd like to remove 
`<%= ' display: none;' unless resource %>` and give it a class instead. Any problems with that?

Also this is what the screen is like if an object accepts two files:
screen shot 2015-05-11 at 5 26 00 pm

parndt:
Yep all looks good to me. In the meantime, I'll merge this PR and you can open another one with a fresh branch.

PR #2979, opened 2015-05-12
Make the functioning of resource-picker more explicit 

anitagraham:
Make the functioning of resource-picker more explicit in code and user view
1. In `admin.js.erb` `init_resources_submit` add the sha parameter to the url returned
   to the caller (resource_picker). Otherwise downloading of a just selected resource fails.
2. Rework to allow for explicit resource replacement
   The view/user should now be in either 
   a. the state where no resource is attached to the
   parent object and they can _add_ a resource. (Only the Add icon/message is visible)
   or 
   b. the state where a resource has been attached to the parent object. 
   Options to _replace, detach or download_ the resource are available.
3. Simplify EN strings, use terms attach/detach/remove/this file
4. Set up skeleton spec for resource_picker.
   (I haven't quite worked out how to do the initial setup.)
5.  Define a replace_icon for use with resource_picker

Add only shows if there is no resource attached

Resource mnaagement icons only appear if there is a resource to manage.

### DIFF
--- a/core/app/assets/javascripts/refinery/admin.js.erb
+++ b/core/app/assets/javascripts/refinery/admin.js.erb
@@ -424,7 +424,7 @@ var link_dialog = {
       e.preventDefault();
       if((resource_selected = $('#existing_resource_area_content ul li.linked a')).length > 0) {
         resourceUrl = parseURL(resource_selected.attr('href'));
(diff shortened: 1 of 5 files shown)


parndt:
This is quite awesome, but I fear it without tests

bricesanchez:
@anitagraham Could you update this awesome PR ?

parndt:
@anitagraham has this gone stale? Should we close it? 😦

PR #2981, opened 2015-05-13
Fixing SQL Server Order By Error

### LINKED ISSUE #2980: Pages admin throws ActiveRecord::StatementInvalid when using SQL Server
Seems to be caused by the 'lft' column being included in the order by twice.  Sql Server doesn't like this.  I can fix it by just removing [this line of code](https://github.com/refinery/refinerycms/blob/master/pages/app/controllers/refinery/admin/pages_controller.rb#L7)  Is that sort necessary here?

```
ActionView::Template::Error (TinyTds::Error: A column has been specified more than once in the 
order by list. Columns in the order by list must be unique.: EXEC sp_executesql N'SELECT [refin
ery_pages].* FROM [refinery_pages] WHERE [refinery_pages].[parent_id] IS NULL ORDER BY lft ASC,
 [refinery_pages].[lft] ASC'):
    1: <ul id='sortable_list'>
    2:   <%= render :partial => 'page', :collection => @pages.roots  %>
    3: </ul>
    4: <%= render '/refinery/admin/sortable_list', :continue_reordering => !!local_assigns[:con
tinue_reordering] %>
```

BrandonJoyce356:
WIP - Need to determine if this breaks anything else, and maybe write a regression test if possible.  
Resolves #2980

### DIFF
--- a/pages/app/controllers/refinery/admin/pages_controller.rb
+++ b/pages/app/controllers/refinery/admin/pages_controller.rb
@@ -4,7 +4,6 @@ module Refinery
       include Pages::InstanceMethods
 
       crudify :'refinery/page',
-              :order => "lft ASC",
               :include => [:translations, :children],
               :paging => false
 
--- a/pages/spec/features/refinery/admin/pages_spec.rb
+++ b/pages/spec/features/refinery/admin/pages_spec.rb
@@ -47,7 +47,8 @@ module Refinery
         end
 
         context "when some pages exist" do
-          before { 2.times { |i| Page.create :title => "Page #{i}" } }
+          let!(:page_0) { Page.create! title: 'Page 0' }
+          let!(:page_1) { Page.create! title: 'Page 1' }
 
           it "shows reorder pages link" do
             visit refinery.admin_pages_path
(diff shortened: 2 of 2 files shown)


parndt:
All tests seem to pass, though I'm not sure if there are any that check that it's ordered properly.

BrandonJoyce356:
I could add one if you want.  Maybe an additional test here that uses the reorder function and then refreshes the page?  https://github.com/refinery/refinerycms/blob/master/pages/spec/features/refinery/admin/pages_spec.rb#L52

parndt:
That sounds very lovely thank you

BrandonJoyce356:
Simulating the drag-n-drop is proving to be quite difficult.  I even tried [this](https://github.com/mattheworiordan/jquery.simulate.drag-sortable.js), but it seems to not work consistently.  Any tricks up your sleeve, or should I just resort to some sort of API call to do the reorder rather than trying to simulate browser interactions?

parndt:
I think calling `update_positions` is best.. but if that's also too troublesome then I don't want to waste your time and we can probably avoid the tests.. we've gotten this far without them! However, tests are the preferred.

BrandonJoyce356:
Let me explore that route for a bit unless you're in a hurry to merge this.  I could work on the tests on a separate PR too.

BrandonJoyce356:
Sorry for the delay on this.  I'll have some more time on it this weekend.

parndt:
@BrandonJoyce356 thanks!

BrandonJoyce356:
This fails when running tests with postgresql or mysql for some reason.  Looking into it.

BrandonJoyce356:
Ah, sqlite starts with index 1, pg starts with 0.  Pretty sure that's it.  I'll find a way to avoid the index numbers.

BrandonJoyce356:
Yay, tests pass.  Ready for review @parndt

parndt:
Works for me!

parndt:
Thanks!

PR #2985, opened 2015-05-25
Use strong params on generated settings_controller

anitagraham:


core/lib/generators/refinery/form/templates/app/controllers/refinery/namespace/admin/settings_controller.rb.erb
```diff
@@ -25,6 +25,11 @@ module Refinery
         end
 
       protected
+
+        def setting_params
+           params.require(:setting).permit(:value)
```
parndt: the indentation is a bit off here

parndt:
Merged, thanks!

PR #2987, opened 2015-05-26
ApplicationHelper does not work with existing Rails 4.2.1 application

rhymes:
I get "uninitialized constant Refinery::AdminController::ApplicationHelper (NameError)" when I try to run any rails commands adding refinerycms to an existing application.

Where that ApplicationHelper comes from?

### DIFF
--- a/core/app/controllers/refinery/admin_controller.rb
+++ b/core/app/controllers/refinery/admin_controller.rb
@@ -3,7 +3,6 @@
 module Refinery
   class AdminController < ::ActionController::Base
     include ::Refinery::ApplicationController
-    helper ApplicationHelper
     helper Refinery::Core::Engine.helpers
     include Refinery::Admin::BaseController
 


parndt:
You don't have one in app/helpers/application_helper.rb ?

rhymes:
Ah ok, that's it! I'm not using helpers in this Rails app so there's no ApplicationHelper anywhere.

parndt:
I think this patch is good but instead:

    helper ::ApplicationHelper if defined?(::ApplicationHelper)

rhymes:
Yes, definitely better. A general question: why do you need to include the ApplicationHelper at all?

parndt:
You know, that's a great question.  It was introduced in 0f16e538

rhymes:
True, but someone else had the same problem in 2012 https://github.com/refinery/refinerycms/commit/0f16e538f6671561f4bdc1f276f68de66dd1daa0#commitcomment-1682559

BTW the tests pass without it

parndt:
Funny that I proposed almost the same solution in 2012.  Thanks for your feedback, I agree with you!

gwagener:
Just updated to the latest Refinery and was caught out by `ApplicationHelper` no longer being included. Good thing the change was so recent so it didn't take long to track down!

parndt:
What does it break??

gwagener:
Custom code added to the Pages admin that relied on a helper defined in `ApplicationHelper`. Pretty easy to fix: there was already a decorator for `Admin::PagesController` so I just moved the helper method there.

parndt:
Ah okay, cool.  Well hopefully others can learn from your example as well!

PR #2988, opened 2015-05-27
Make layout and view template patterns configurable

anitagraham:
Tests ran successfully locally.
Tests for this feature not yet written


parndt:
what's the example use case?

anitagraham:
I use .rb templates with mustache. 
They don't have 'html' in their name and so they are not found and shown in _form_advanced_options.

parndt:
Well, that seems fair enough.. perhaps the tests could focus around that? :smile: thanks

PR #2994, opened 2015-06-04
Getting Started guide: added notes about creating a Rails 4 application

r11runner:
Hi,

I've just started to play with Refinery CMS, and for me, it wasn't obvious how I can create a Rails 4 application with Refinery. So I thought that I'll update the Getting Started guide with my "discoveries". :) If they're wrong, please let me know what would be the correct way.
Thanks.

Cheers,
r11runner


parndt:
Nice one :+1:

PR #3004, opened 2015-07-01
Update _google_analytics.html.erb

krisquigley:
Updated ga.js to newer analytics.js


bricesanchez:
It's ok for me and you @parndt ?

PR #3008, opened 2015-07-18
Remove dragonfly_insert_before, as it's never used.

### LINKED ISSUE #3000: Refinery::Images.config.dragonfly_insert_before
I can't see that this parameter is used. Is it a leftover from the past or a promise for the future?

parndt:
Fixes #3000


simi:
:question: why was this added?

parndt:
No idea, but it was years ago and was never used, so it's removed.

PR #2999, opened 2015-06-17
New revised user experience - Glass

stefanspicer:
On behalf of my entire team at Glass Canvas, I'm submitting this Pull Request.  It is a start towards moving Refinery to the next level.  Refinery is a great code base and an amazing CMS, and we'd like to see it make splash in the world of CMSs!  Who's with us?  At Glass Canvas, we work solely with non-profit and social efforts to give them the best, and refinery is the best!

Development of 'glass' (its code name that you'll see all over the code) started in Oct 2014 and after almost 1000 commits (that I've condensed into a few) we've launched this new refinery on 4 production websites.  This addresses much of what is discussed in issues #2731 #2934 #2833 and others.  I just spent a few days and integrated this with master, as it was based off of 2.1.5.  Our production websites are still running 2.1.5.

## **Main** Features of Refinery Glass
* New look - main menu, sidebars, icons, contextual menu (#actions), responsive
* glass editor to replace wymeditor with grande.js integrated with refinery - edit content in place as it will show live
* Ajax is used for forms to enhance the user experience
* Creating a new admin sends an invite email so they can activate their account and pick a password.  This makes for a more professional "Welcome to your CMS" experience than telling them their username and password.
 
Here is a quick demo I created for @parndt a few months ago. http://youtu.be/Gh1l_Yh8Mvo  Download and give it a spin, I'm sure people will have questions and comments.

## Still a lot left to do
* I've marked a bunch of tests as skipped, the reason shows my explanation.  This is a good place to start to get things up to speed
* The editor is very simple, doesn't have some of the features of wymeditor
* Multiple locales on pages needs to be re-integrated into the interface, and re-thought through
* I have a list that we need to prioritize, the above are the biggest things.

PR #3009, opened 2015-07-19
Merge cjbrew forms

parndt:
Replaces #2922


parndt:
@simi what do you think about these changes, given your work on generators lately?

simi:
Hello @parndt. Give me please day or two, since this is a huge change. It's honest to review this. :heart:

stefanspicer:
I have not been following all the refinery issues and PR's as well as I'd like, but I'm really trying to stay in the loop more.  This is my first time looking at this on.  What, in a sentence, is this PR for?

I see:

* improved specs (if there were any there before?)
* includes_spam option

In #2999 I think I may have duplicated some of this work.  I've found the "moderation" (from the blog engine) feature covers the "spam" feature and more, so I integrated the blog moderation into the form generator.  I then found that every second time I generated a form I needed the objects to be displayed on the frontend of the site too (a testimonial, a comment, a story), so after doing this a few times I integrated the form generator into the engine generator.  I thought it made sense for them to be one.

Thoughts on this approach?

parndt:
@stefanspicer Doesn't the engine generator already generate frontend? (index/show) whereas the form generator generates forms that get submitted on the frontend.
So, the engine generator handles the 99% of extensions which require no user input on the frontend except to look at things.

parndt:
Either way you look at it, all of our generators generate far too much code and are too complicated to maintain.

stefanspicer:
Yes, the engine generator generates frontend.  The problem I ran into was that I often needed a form (for a comment, testimony, story) that needed these features:

* [form] frontend form
* [form] "moderation" - reject / approve
* [form] notification email & setting for who to notify
* [engine] admin add/edit the comments
* [engine] frontend index view
* [engine] admin reorder (one case needed this)

I do realize that these are all fairly easy just to copy and paste in, but it just made sense to me that if these features are in any generator, then they should all be in the same generator.  That's why I merged the form and engine generator.  It is more complex, but I think it is less code to maintain overall.

I've never actually found that the generators generate too much code.  I do understand your pain in maintaining them!  .erb.erb in itself gets very messy.

bricesanchez:
Could we milestone it to 3.0 ?

bricesanchez:
:tada:

bricesanchez:
@CJBrew Thank you for your awesome job on this PR !

PR #3011, opened 2015-07-20
Fix #3010 page parts

### LINKED ISSUE #3010: Page parts title missing on newly updated website
Since the merge of #2875, if you update an existing website on refinerycms 3.0 master "hide_sections" option on render_content_page doesn't works.

It's because `hide_sections` tries to hide a part title and not a part slug.

i'm searching how to fix that. There is no data lose.

bricesanchez:
Fix for #3010 

@parndt

### DIFF
--- a/pages/db/migrate/20150720155305_update_slug_and_title_in_refinery_page_parts.rb
+++ b/pages/db/migrate/20150720155305_update_slug_and_title_in_refinery_page_parts.rb
@@ -1,0 +1,11 @@
+class UpdateSlugAndTitleInRefineryPageParts < ActiveRecord::Migration
+  def change
+    Refinery::PagePart.all.each do |pp|
+      unless pp.title
+        pp.title = pp.slug
+      end
+      pp.slug = pp.slug.downcase.gsub(" ", "_")
+      pp.save!
+    end
+  end
+end
--- a/pages/db/migrate/20150720155305_update_slug_and_title_in_refinery_page_parts.rb
+++ b/pages/db/migrate/20150720155305_update_slug_and_title_in_refinery_page_parts.rb
@@ -1,11 +1,13 @@
 class UpdateSlugAndTitleInRefineryPageParts < ActiveRecord::Migration
   def change
-    Refinery::PagePart.all.each do |pp|
-      unless pp.title
-        pp.title = pp.slug
+    begin
+      ::Refinery::PagePart.all.each do |pp|
+        pp.title ||= pp.slug
+        pp.slug = pp.slug.downcase.gsub(" ", "_")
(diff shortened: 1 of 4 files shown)


pages/db/migrate/20150720155305_update_slug_and_title_in_refinery_page_parts.rb
```diff
@@ -0,0 +1,11 @@
+class UpdateSlugAndTitleInRefineryPageParts < ActiveRecord::Migration
+  def change
+    Refinery::PagePart.all.each do |pp|
+      unless pp.title
+        pp.title = pp.slug
+      end
+      pp.slug = pp.slug.downcase.gsub(" ", "_")
+      pp.save!
+    end
```
parndt: It is generally not considered acceptable to change data in a migration.. is there another way that this can be handled for the user?

ghoppe:
Perhaps a rake task?

pages/db/migrate/20150720155305_update_slug_and_title_in_refinery_page_parts.rb
ghoppe: Perhaps a rake task?

bricesanchez: it has to be mandatory for the user, i don't know to do it in other way. Do you have any solution?

bricesanchez:
@parndt,  I've duplicate and run the migration 3 times without errors.

pages/db/migrate/20150720155305_update_slug_and_title_in_refinery_page_parts.rb
```diff
@@ -0,0 +1,11 @@
+class UpdateSlugAndTitleInRefineryPageParts < ActiveRecord::Migration
+  def change
+    if defined?(::Refinery::PagePart)
+      Refinery::PagePart.all.each do |pp|
+        pp.title ||= pp.slug
+        pp.slug = pp.slug.downcase.gsub(" ", "_")
+        pp.save!
+      end
+    end
+  end
```
parndt: Actually, this won't work.  Can you use the style from https://github.com/refinery/refinerycms/commit/8f56bf47a7ecd6d348b26d9531ff5e4eb03d5f68 where we use begin + rescue instead? This ensures that the classes at least try to be loaded before declaring that they don't exist as `defined?` doesn't try.

bricesanchez: I'm doing this now. Don't go too far ;)

PR #3015, opened 2015-07-22
Removed flash messages from admin dialogues.

### LINKED ISSUE #2995: Flash message duplicated in dialog
1.  Edit page content and `save and continue`
A flash message "`page_name` was successfully updated" appears.
2. Without closing the flash container select "Add image" or "Add link".
A dialog opens with the same flash message at the top. The flash container obscures part of the dialog content.
flash message in dialog box

Possibly the flash containers need ids to distinguish one from the other. 
If the flash container is visible it shouldn't obscure other content.

I will look in more detail when time permits.

parndt:
Fixes #2995


bricesanchez:
Thanks @parndt !

anitagraham:
Thanks @parndt

PR #3017, opened 2015-07-23
Add dragonfly_verify_urls as a config option for images and resources

anitagraham:


images/spec/models/refinery/image_spec.rb
```diff
@@ -98,6 +98,21 @@ module Refinery
...
+        end
+      end
+
+      context "when Dragonfly.verify_urls is false" do
+        Refinery::Images.dragonfly_verify_urls = false
+        it "returns a url without an SHA parameter" do
+          expect(created_image.url).not_to match(/\?sha=[\da-fA-F]{16}\z/)
+        end
+      end
+
```
parndt: context "when Dragonfly.verify_urls is false" do
  before { allow(Refinery::Images).to receive(:dragonfly_verify_urls).and_return(false) }
  it "returns a url without an SHA parameter" do
    expect(created_image.url).not_to match(/\?sha=[\da-fA-F]{16}\z/)
  end
end

images/spec/models/refinery/image_spec.rb
```diff
@@ -98,6 +98,21 @@ module Refinery
       it "can resize and strip a thumbnail" do
         expect(created_image.thumbnail(:geometry => '200x200', :strip => true).url.blank?).to eq(false)
       end
+
+      context "when Dragonfly.verify_urls is true" do
+        Refinery::Images.dragonfly_verify_urls = true
+        it 'has an SHA parameter at the end of the URL' do
+          expect(created_image.url).to match(/\?sha=[\da-fA-F]{16}\z/)
+        end
+      end
```
parndt:
```ruby
context "when Dragonfly.verify_urls is true" do
  before { allow(Refinery::Images).to receive(:dragonfly_verify_urls).and_return(true) }
  it 'has an SHA parameter at the end of the URL' do
    expect(created_image.url).to match(/\?sha=[\da-fA-F]{16}\z/)
  end
end
```

anitagraham:
Thanks for the `allow/receive` syntax @parndt . 
The test for `when Dragonfly.verify_urls is false`  still fails. 
It seems to me that this is because each Dragonfly app is initialized once and the tests run as a single instance.

An alternative (not very attractive) is to test `verify_urls=true` in Refinery::Images and `verify_urls=false` in Refinery::Resources as these are separate Dragonfly apps.

anitagraham:
so this works:

````ruby
before {
          allow(Refinery::Images).to receive(:dragonfly_verify_urls).and_return(false)
           ::Refinery::Images::Dragonfly.configure!
        }
````

**Comment by parndt** [comment]
That makes sense!

File: resources/spec/models/refinery/resource_spec.rb [hunk 3]
```diff
@@ -22,6 +22,28 @@ module Refinery
...
+        it "returns a url with an SHA parameter" do
+          expect(resource.url).to match(/\?sha=[\da-fA-F]{16}\z/)
+        end
+      end
+
+      context "when Dragonfly.verify_urls is false" do
+        before {
+          allow(Refinery::Resources).to receive(:dragonfly_verify_urls).and_return(false)
+          ::Refinery::Resources::Dragonfly.configure!
+        }
```
parndt: now that the before case is multi line we should switch to a block:

```ruby
before do
  allow(Refinery::Resources).to receive(:dragonfly_verify_urls).and_return(false)
  ::Refinery::Resources::Dragonfly.configure!
end
```

parndt:
Seems like the tests are failing on CI?

parndt:
Strange, I ran them again and they don't fail now.
I'll rebuild the entire thing a few times just to make sure.

parndt:
Seems okay, so with the `{ }` -> `do end` change we should be good to go..

bricesanchez:
@parndt Are you ok to merge it ?

parndt:
@bricesanchez yep I'll get it merged

anitagraham:
The tests for dragonfly_verify_urls are failing intermittently. I have been considering them and other dragonfly/images stuff I am looking at - it would be better to test that config values passed to dragonfly by Refinery are reflected by that dragonfly app.

Q:  Should I remove *these* specific tests while I write the new tests?

parndt:
@anitagraham I'm really curious as to why they're failing intermittently?

anitagraham:
I wonder if  the reconfiguration is taking effect?
Intermittent might then depend on test order.
If I can define a two dragonfly apps with different configs and get different (desired) results then it might be a better test.

I have written a separate refinerycms-dragonfly 'gem' and I am up to writing tests for that anyway.
It would make adding other dragonfly apps easier, and it would, in the long run, allow refinery to use different image handlers (someone mentioned cloudinary earlier this year.)

I had done the work anyway for a refinery engine which uses two dragonfly apps to keep images of vastly different types separate from Refinery::Images and from each other.

parndt:
Wow, It'd be great to remove all of the dragonfly coupling from the core!

PR #3019, opened 2015-07-25
Update 4 - Pretty URLs.textile

mlinfoot:
Works with Refinery 3, but I've not done any testing.

No need to manually create the migration -- use rails generate migration ...

Added "Refinery" where needed (eg refinery_staff_members

### DIFF
--- a/doc/guides/4 - Refinery Extensions/4 - Pretty URLs.textile
+++ b/doc/guides/4 - Refinery Extensions/4 - Pretty URLs.textile
@@ -8,30 +8,31 @@ This guide will show you how to:
 
 endprologue.
 
-WARNING: This method is tested with Refinery CMS v. 2.0.0 only. It should be applicable to later versions, but not earlier versions (1.0.9 or earlier). Always ensure your code is backed up or under version control before making substantial change.
+WARNING: This method is tested with Refinery CMS v. 2.0.0 only. It should be applicable to later versions, but not earlier versions (1.0.9 or earlier). It also works with v 3.0, although it has not been tested extensively.
+Always ensure your code is backed up or under version control before making substantial changes.
 
 h3. Fresh Slugs
 
 FriendlyID requires that you create a string field entitled +slug+ on any model that you wish to convert. It's better to do this as a separate migration after creating the extension, because you don't want Refinery to auto-generate a field in the back-end; this attribute is set by code, not by hand.
 
-To do so, create a new migration file in the +db/migrate+ folder of your extension. Since this depends on the previous migration, it's best to rename the other migrations to include a sequence reference in the filename -- for example, +create_staff_members.rb+ becomes +001_create_staff_members.rb+. When you re-run the generator that will move these migrations over to your main +db/migrate+ folder, they will be moved in this order. Don't worry if you've already run these migrations; Rails will auto-detect that you've run them, even though the filename has been changed, and it won't re-run them.
+To do so, from the root of your app, generate a new migration file to add the +slug+ column to your extensions's database table. We'll use a StaffMembers extension as an example.
+<shell>
+$ rails generate migration AddSlugToRefineryStaffMembers slug:string:index
+</shell>
 
-Inside your migration, write something like this:
+Your migration, +/db/migrate/..._add_slug_to_refinery_staff_members+ should look like this:
 <ruby>
-  class AddSlugToStaffMembers < ActiveRecord::Migration
+  class AddSlugToRefineryStaffMembers < ActiveRecord::Migration
     def change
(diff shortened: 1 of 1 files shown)


doc/guides/4 - Refinery Extensions/4 - Pretty URLs.textile
```diff
@@ -8,30 +8,31 @@ This guide will show you how to:
...
-      add_index :staff_members, :slug
+      add_column :refinery_staff_members, :slug, :string
+      add_index :refinery_staff_members, :slug
     end
   end
 </ruby>
 
-Run the generator that you ran initially to create your extension, and then migrate. In the above example, since the extension is named +staff_members+, one would run:
-
+Now run
```
parndt: I think this is missing a colon

`Now run:`

doc/guides/4 - Refinery Extensions/4 - Pretty URLs.textile
```diff
@@ -8,30 +8,31 @@ This guide will show you how to:
...
+Always ensure your code is backed up or under version control before making substantial changes.
 
 h3. Fresh Slugs
 
 FriendlyID requires that you create a string field entitled +slug+ on any model that you wish to convert. It's better to do this as a separate migration after creating the extension, because you don't want Refinery to auto-generate a field in the back-end; this attribute is set by code, not by hand.
 
-To do so, create a new migration file in the +db/migrate+ folder of your extension. Since this depends on the previous migration, it's best to rename the other migrations to include a sequence reference in the filename -- for example, +create_staff_members.rb+ becomes +001_create_staff_members.rb+. When you re-run the generator that will move these migrations over to your main +db/migrate+ folder, they will be moved in this order. Don't worry if you've already run these migrations; Rails will auto-detect that you've run them, even though the filename has been changed, and it won't re-run them.
```
parndt: I prefer this older instruction.  It's a good idea to place the migration inside the extension and then using `rails g refinery:staff_members` to copy it to the application.  This ensures that if you gemify the extension that it has everything that it needs right out of the box.  Otherwise, if you copy this to another application, it will be missing some of its migrations.

mlinfoot: Thanks for looking at this Philip -- you see the big picture. I was thinking in terms of an extension that was very specific to my app, and in that context, couldn't see why anyone would want/need to create the migration by hand and then take the extra step to copy it to the application. I hope my other suggestions are more helpful.

doc/guides/4 - Refinery Extensions/4 - Pretty URLs.textile
mlinfoot: Yes, thank you!

doc/guides/4 - Refinery Extensions/4 - Pretty URLs.textile
parndt: Yep, I like the rest, but I want to keep the migration advice the same. :smile:

mlinfoot: Glad to help a bit where I can.

On Sun, 26 Jul 2015 at 20:57 Philip Arndt <notifications@github.com> wrote:

> In doc/guides/4 - Refinery Extensions/4 - Pretty URLs.textile
> <https://github.com/refinery/refinerycms/pull/3019#discussion_r35500362>:
>
> >
> >  h3. Fresh Slugs
> >
> >  FriendlyID requires that you create a string field entitled +slug+ on any model that you wish to convert. It's better to do this as a separate migration after creating the extension, because you don't want Refinery to auto-generate a field in the back-end; this attribute is set by code, not by hand.
> >
> > -To do so, create a new migration file in the +db/migrate+ folder of your extension. Since this depends on the previous migration, it's best to rename the other migrations to include a sequence reference in the filename -- for example, +create_staff_members.rb+ becomes +001_create_staff_members.rb+. When you re-run the generator that will move these migrations over to your main +db/migrate+ folder, they will be moved in this order. Don't worry if you've already run these migrations; Rails will auto-detect that you've run them, even though the filename has been changed, and it won't re-run them.
>
> Yep, I like the rest, but I want to keep the migration advice the same. [image:
> :smile:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/refinery/refinerycms/pull/3019/files#r35500362>.
>
-- 

marshal

parndt:
@mlinfoot do you intend to make the changes? :smile:

mlinfoot:
@parndt Goes to show you how clueless I am. :( I think I get it now and, yes, I will make the changes and send another pull request.

PR #3022, opened 2015-07-30
Pretty URLs

mlinfoot:
Added that it works with RefineryCMS 3.0. 
Changed "staff_members" to "refinery_staff_members" in migrate example.


parndt:
Great, thank you

mlinfoot:
Glad I could give a little back.

PR #3027, opened 2015-08-03
Fix issue 3025

anitagraham:
1. call translation string with scope
2. Add class ‘icon’ to page branches

Class 'icon' isn't the *semantic* class for this, however it is what the javascript in `tree.js` and `sortable_list.js` are looking for.  As I'm not sure who else uses these it's probably unsafe to change `.icon` into `.branch` or `.parent`.

### DIFF
--- a/pages/app/views/refinery/admin/pages/_page.html.erb
+++ b/pages/app/views/refinery/admin/pages/_page.html.erb
@@ -11,7 +11,9 @@
   <li class='clearfix record items' id="<%= dom_id(page) -%>">
   <div class='clearfix'>
     <% if page.children.present? %>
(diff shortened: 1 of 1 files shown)


anitagraham:
Hmmm... the fails I've checked are `dragonfly_verify_url` fails.
Obviously that test isn't as good as it should be.

parndt:
@anitagraham can you please fix that test up ASAP? Otherwise, I'll have to revert the feature in order to get CI green on master again :anguished:

anitagraham:
[Comments on testing dragonfly_verify_urls](https://github.com/refinery/refinerycms/pull/3017)

anitagraham:
PR #3031

PR #3031, opened 2015-08-08
Test Dragonfly configuration separately

anitagraham:
Always call Dragonfly.app.verify_urls
Test Refinery->Dragonfly configuration
Remove intermittently failing test.


parndt:
Ace!

PR #3038, opened 2015-08-18
Improve menu presenter

bricesanchez:
Now there is a difference between an active item and a current item.
Useful for nested navigation.


simi:
@bricesanchez is there any chance to keep this in upgrade guides?

bricesanchez:
Or i could keep `selected_css` and only add `active_css` ?

simi:
It will be great to keep old behaviour and add new only.

bricesanchez:
Ok :)

simi:
:-*

PR #3043, opened 2015-08-21
Add methods to permit resource params

bricesanchez:
In order to be able to override this methods in an extension.

Works like pages : 
https://github.com/refinery/refinerycms/blob/master/pages/app/controllers/refinery/admin/pages_controller.rb#L99-L105


simi:
Danke!

PR #3046, opened 2015-08-24
Added plugin ordering set by config option.

### LINKED ISSUE #3024: Cannot reorder admin tabs
Upgrading to latest commit (9a69a74) resulted in no way to reorder admin tabs (Users, Pages, etc.). The reorder icon visible in previous versions no longer appears.

Reproduced with a clean, fresh install --  rails new myapp -m http://refinerycms.com/t/edge

gwagener:
Fixes #3024.

### DIFF
--- a/core/spec/lib/refinery/plugins_spec.rb
+++ b/core/spec/lib/refinery/plugins_spec.rb
@@ -21,6 +21,33 @@ module Refinery
     end
   end
 
+  module MyOrderedPlugin
+    class Engine < ::Rails::Engine
+      isolate_namespace ::Refinery
+      ::Refinery::Plugin.register do |plugin|
+        plugin.name = "my_ordered_plugin"
+      end
+    end
+  end
+
+  module MyOtherOrderedPlugin
+    class Engine < ::Rails::Engine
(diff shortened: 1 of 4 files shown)


core/spec/lib/refinery/plugins_spec.rb
```diff
@@ -61,6 +97,11 @@ class Engine < ::Rails::Engine
         expect(subject.registered.in_menu.any?).to be_truthy
         expect(subject.registered.in_menu.all? { |p| !p.hide_from_menu }).to be_truthy
       end
+
+      it "orders by plugin_priority config" do
+        Core.config.plugin_priority = %w(my_ordered_plugin my_plugin my_other_ordered_plugin)
+        expect(subject.registered.in_menu.names.take(2)).to eq %w(my_ordered_plugin my_other_ordered_plugin)
```
parndt: I think that to avoid the new routing errors we've just seen in CI, we are going to need to mock this.

Something like:

```ruby
allow(subject).to receive(:registered).and_return([
  double(name: "one"), 
  double(name: "two")
])
```

Or something like that.

parndt:
See #3053

PR #3059, opened 2015-09-30
Add some missing Dutch translations and a generic one.

johanb:
Fixed some missing translations in the Dutch language for the title & alt attributes of images.

Also noticed the missing "locale_picker.language" in English and German, so fixed that to.


johanb:
Added Spanish for bonus points ;)

bricesanchez:
Thanks !

PR #3062, opened 2015-10-20
Add new guide for upgrading to Refinery 3.0

anitagraham:
As requested, and with thanks to stefanspicer and charlesdeb for modifications.


bricesanchez:
Thanks @anitagraham !

PR #3068, opened 2015-10-29
Set hide_cancel to false for existing_resource

bricesanchez:
It fixes the close action of the dialog in the https://github.com/anitagraham/refinerycms-page-resources extension. 

See : https://github.com/anitagraham/refinerycms-page-resources/pull/1


bricesanchez:
@anitagraham What do you think of this PR ?

anitagraham:
Looks good. Does it have a test?

bricesanchez:
No, i think it's more an hotfix for this problem than a real solution.

PR #3074, opened 2015-11-30
Add page_options

anitagraham:
Visiting an old topic with a different approach, I think.  Allowing extensions to add their own fields to the Page advanced options form. The current option of using a partial `_form_extra_fields_for_more_options.html.erb` may cause conflicts between different engines.
One might do this when the extension adds fields to the page using a decorator.

Add a page_option which will allow extensions to add extra options to
_form_advanced_options.  
**Example**

``` ruby
module Refinery
  module Myextension
    class Engine < Rails::Engine

      def self.register_page_options(options)
        options.name = 'myextension'
        options.partial = '/refinery/myextension/admin/pages/page_options'
      end
......

      config.after_initialize do
        Refinery.register_extension(Refinery::Myextension)
        Refinery::Pages::PageOptions.register do |option|
          register_page_options option
        end
      end
```

The partial in `/refinery/myextension/admin/pages/page_options` would appear in the Advanced Options section of the Page edit form.

### DIFF
--- a/pages/app/views/refinery/admin/pages/_form_advanced_options.html.erb
+++ b/pages/app/views/refinery/admin/pages/_form_advanced_options.html.erb
@@ -78,7 +78,9 @@
       <%= f.label :show_in_menu, t('.show_in_menu_description'),
                   :class => "stripped" %>
     </div>
-
+    <% Refinery::Pages.page_options.collect do |options| %>
+      <%=  render partial: options.partial, locals:{f: f} %>
+    <%  end   %>
     <%= render 'form_extra_fields_for_more_options', :f => f %>
   </div>
 
--- a/pages/app/presenters/refinery/pages/menu_presenter.rb
+++ b/pages/app/presenters/refinery/pages/menu_presenter.rb
@@ -1,4 +1,3 @@
-require 'active_support/core_ext/string'
 require 'active_support/configurable'
 require 'action_view/helpers/tag_helper'
 require 'action_view/helpers/url_helper'
--- a/pages/lib/refinery/pages.rb
+++ b/pages/lib/refinery/pages.rb
@@ -8,6 +8,7 @@ module Refinery
     require 'refinery/pages/tab'
     require 'refinery/pages/type'
     require 'refinery/pages/types'
+    require 'refinery/pages/page_options'
 
     # Load configuration last so that everything above is available to it.
     require 'refinery/pages/configuration'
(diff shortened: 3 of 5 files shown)


bricesanchez:
What about using `spree deface` ( https://github.com/spree/deface ) instead of partial files ?

parndt:
The issue I have with this feature is: where does it stop?  If we add a little API here and there for adding more and more things to pages, I'm concerned that it'll become (more) bloated and we will have to support these APIs forever.

anitagraham:
A page in the admin interface consists of a title, a collection of body parts, and a collection of page options. (Even the title is a body part)
So, I think that an admin page rendering should be a series of API calls.
Provide a page wrapper and inside it
   For each body part: render content
   For each 'page control': render the control.

The current advanced options comprise about four or five page controls.
Hmmm... can I do what I want with a decorator for the admin page view?

Brice: I will look at the Spree::Deface too. 
I don't like anything that relies on existing markup for much the same reasons I don't want to user view overrides. 
Thanks for the ideas.

anitagraham:
Deface seems to get a big :-1:  from Spree or Spree users for contributing to unmaintainable/undebuggable code and general confusion.

bricesanchez:
I use Deface for more than 5 projects without any troubles. I prefer this way to add/remove little code blocks in my main app instead of overriding a view. It could be confusing if we use too much this extension.

bricesanchez:
@parndt are you ok with feature ?
If yes, @anitagraham could you add some specs for this feature ?

parndt:
@bricesanchez https://github.com/refinery/refinerycms/pull/3074#issuecomment-162395341

johanb:
I'm a :-1: on Deface. I sympathize with the desire for this feature though. Although it would be unclear which extension certain options come from. Which could potentially become also a maintenance burden.

parndt:
Maybe it is time that we put some thought into how pages should work and how they should interact with the rest of the system ?

bricesanchez:
@parndt yes :)

pages/app/views/refinery/admin/pages/_form_advanced_options.html.erb
```diff
@@ -78,7 +78,9 @@
       <%= f.label :show_in_menu, t('.show_in_menu_description'),
                   :class => "stripped" %>
     </div>
-
+    <% Refinery::Pages.page_options.collect do |options| %>
```
simi: Why `collect`?

bricesanchez:
What should we do with this PR @parndt and @anitagraham ?

IMHO, we should close it and rewrite the frontend.

PR #3078, opened 2015-12-13
Fix broken link

LiamMayfair:
Fix broken link to _Generate an Extension to Use Your Own MVCs_.


parndt:
Thanks @LiamMayfair ! Seems like it also still references `2.1.0` not `3.0.0`

PR #3080, opened 2015-12-17
Lock sprockets-rails to ~> 2.3.3

bricesanchez:
Speakingurl block refinerycms start if you have sprockets-rails 3.0.0.

I've send a PR to this gem : https://github.com/pid/speakingurl/pull/70

PR #3084, opened 2015-12-30
Set speakingurl-rails to 8.0.0, switch to poltergeist gem.

parndt:
Replaces #3080

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gemspec
 gem 'quiet_assets'
 gem 'spring'
 gem 'spring-commands-rspec'
-gem 'poltergeist', github: 'teampoltergeist/poltergeist'
+gem 'poltergeist', '>= 1.8.1'
 
 # Add support for refinerycms-acts-as-indexed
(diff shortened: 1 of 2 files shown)


parndt:
@bricesanchez the sprockets error has disappeared, but now `getSlug` is undefined in the javascript.

parndt:
@bricesanchez this error is because `speakingurl.js` is not found

parndt:
I have to say this is quite a large dependency for such a small feature (used in one place, optionally)

bricesanchez:
Yeah you're right, that's not good for us

bricesanchez:
I think i've the fix https://travis-ci.org/refinery/refinerycms/builds/100263612
I've posted a new PR to speakingurl-rails : https://github.com/pid/speakingurl/pull/71
We should wait for his version bump

parndt:
what about apps that aren't using sprockets? Must they now use it?

bricesanchez:
I'm a little confused on what is the best way to fix this issue : 
We could just add the js directly in our assets directory ?

bricesanchez:
I've merged this PR because it's fixing the sprocket-rails 3.0 blocker issue.

PR #3089, opened 2016-01-23
parse image url to match html as written

anitagraham:


bricesanchez:
Thanks a lot @anitagraham !

PR #3092, opened 2016-02-01
Specify that we are expecting action_name to be insert

parndt:
This prevents a crash issue where it tries to access attributes on the form (which is nil) because the action_name was resolving as create.


bricesanchez:
Thanks !

PR #3093, opened 2016-02-01
Add line numbers to stack trace

jaredbeck:
If one does not explicitly add line numbers to an eval, they will
be lacking in stack traces, and in the output of
`Method#source_location`.

There are many other evals in refinery, but they use the zero-
-argument method signature (the signature that takes a block), and
I'm not sure if they need explicit line numbers.

## References

- http://railsguides.net/find-method-source-location/
- https://ernie.io/2012/09/10/ruby-tidbit-__line__-and-heredocs/

### DIFF
--- a/core/lib/refinery/crud.rb
+++ b/core/lib/refinery/crud.rb
@@ -227,7 +227,7 @@ module Refinery
                     :create_or_update_successful,
                     :create_or_update_unsuccessful,
                     :merge_position_into_params!
-        )
+        ), __FILE__, __LINE__
 
         # Methods that are only included when this controller is searchable.
         if options[:searchable]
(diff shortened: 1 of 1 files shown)


jaredbeck:
23 of 24 build jobs pass. The one that fails seems unrelated, so I'd recommend retrying that job (I don't seem to have permission to retry)

bricesanchez:
@jaredbeck done.

parndt:
@jaredbeck this is awesome! I'd prefer that we didn't eval at all but it's so much time to move to a better solution so this is a great help. Thank you very much!

jaredbeck:
> this is awesome!

:) Glad you like. I just learned about it.

> I'd prefer that we didn't eval at all ..

Haha yeah it messes up syntax highlighting too :) (in my editor anyway)

> .. it's so much time to move to a better solution .. [without eval]

I bet!

PR #3101, opened 2016-02-09
Add protect_from_forgery with: :exception

bricesanchez:


simi:
Are you sure this will fix it? No specs?

bricesanchez:
i don't know how to test it and i asking myself why we added this line in core/lib/refinery/application_controller.rb but it fix the CSRF discovered by @cor3sm4sh3r.

simi:
I saw that. But this just raise exception instead of using blank session when CSRF is detected. Am I right? I'm not sure how it can solve this problem.

simi:
Si am I not right?

parndt:
@simi I thought that it was better to raise an exception to prevent the possibility of anything being created/used incorrectly by the system.

simi:
You're out of scope of creating or using things since you have empty session for that request.

simi:
I'm not against this patch, but I don't think it is fixing any issue.

parndt:
Fair point.  At least by handling this with an exception it gives developers an opportunity to rescue this exception and do something else?

cor3sm4sh3r:
hi all,
Raising the exception fixed csrf issue but i think this should be more
likely solution.

rescue_from ActionController::InvalidAuthenticityToken do |exception|
  sign_out_user # Example method that will destroy the user cookies
end

On Fri, Feb 12, 2016 at 3:26 AM, Philip Arndt <notifications@github.com>
wrote:

> Fair point. At least by handling this with an exception it gives
> developers an opportunity to rescue this exception and do something else?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/refinery/refinerycms/pull/3101#issuecomment-183075143>
> .
>



-- 
shravan kumar budharap
cor3sm4sh3r

parndt:
@simi @bricesanchez  ^ I think that suggested code is better too, what do you think?

bricesanchez:
:+1:

simi:
http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection/ClassMethods.html#method-i-protect_from_forgery `:reset_session` option is there. But as I stated multiple times here, I don't see a reason how this can fix some issue.

The only difference is with this exception is raised, without this empty session is provided for given request (future valid requests can use this session again). There's third option (:reset_session) to destroy session at all (no future valid request can continue in this session).

But in no case you should be able to access session when CSRF protection is activated.

Or am still I missing something?

bricesanchez:
I thinking about creating a spec to check this issue. So after that qe will see if it really works for us but i agree with @simi there is probably something inside this bug we doesn't discover yet...

PR #3104, opened 2016-02-11
Shelly Cloud will shut down on March 31, 2016.

bricesanchez:
https://shellycloud.com/


simi:
Should we wait until it is down?

bricesanchez:
Yes, i milestone it for 3.1.

parndt:
If it's going to shut down then it doesn't make much sense to advise people on how to use it! :smile:

simi:
true

PR #3108, opened 2016-02-12
Use font-awesome from font-awesome-sass gem

ogerman:
That resolves conflict with font-awesome-sass gem (refinerycms has higher asset paths priority, and overrides it)
Also it makes refinerycms gem more lightweight and clean


bricesanchez:
@ogerman Nice!

@anitagraham What do you think of this PR? Is it ok with your icon set feature?

anitagraham:
Yes, it should be OK. I think there was a problem with the font-awesome-sass gem which meant I couldn't use it then. I guess it has been resolved.

bricesanchez:
Thanks @ogerman and @anitagraham !

PR #3124, opened 2016-03-07
Specify the correct new_page_part_params.

### LINKED ISSUE #3110: Creating a new content section results in an "found unpermitted parameters" error
In version 3.0.1. I have turned set "config.new_page_parts = true".  From my "About" page I try to create a new page part by clicking the the add icon.  The dialog comes up where I enter my title and slug.  But when I hit the "Save" button I get the following error:
```
Started GET "/refinery/page_parts/new?title=bottom&part_index=2&slug=bottom&body=" for 127.0.0.1 at 2016-02-13 11:49:42 -0800
Processing by Refinery::Admin::PagePartsController#new as HTML
  Parameters: {"title"=>"bottom", "part_index"=>"2", "slug"=>"bottom", "body"=>"", "locale"=>:en}
Completed 500 Internal Server Error in 2ms (ActiveRecord: 0.0ms)

ActionController::UnpermittedParameters - found unpermitted parameters: part_index, locale:
  actionpack (4.2.5.1) lib/action_controller/metal/strong_parameters.rb:497:in `unpermitted_parameters!'
  actionpack (4.2.5.1) lib/action_controller/metal/strong_parameters.rb:339:in `permit'
   () usr/local/rvm/gems/ruby-2.1.2/bundler/gems/refinerycms-019f18356d6a/pages/app/controllers/refinery/admin/page_parts_controller.rb:26:in `new_page_part_params'
   () usr/local/rvm/gems/ruby-2.1.2/bundler/gems/refinerycms-019f18356d6a/pages/app/controllers/refinery/admin/page_parts_controller.rb:7:in `new'
```

The problem is that the "Save" button is resulting in a request to the new action of the PagePartsController.   In the admin.js.erb file in the core refinery gem I see that the Save button has a click event handler assigned to it on line 606.  That function is doing an ajax GET to the new page part url.  It seems like this should be sending the request to a create action in PagePartsController but there is only a new and a destroy action.  So not sure what the fix should be here.  Shouldn't the event handler callback function be doing a POST to a create action in PagePartsController?

parndt:
Allow locale, which is reasonable.
Exclude part_index which is used internally

Fixes #3110

### DIFF
--- a/pages/app/controllers/refinery/admin/page_parts_controller.rb
+++ b/pages/app/controllers/refinery/admin/page_parts_controller.rb
@@ -23,7 +23,7 @@ module Refinery
 
       protected
         def new_page_part_params
-          params.permit(:title, :slug, :body)
+          params.except(:part_index).permit(:title, :slug, :body, :locale)
         end
 
     end


bricesanchez:
Thanks!

PR #3125, opened 2016-03-08
Add TranslatedFieldPresenter

### LINKED ISSUE #2870: Refactor page_title_with_translations method
Hello !

I would like to make this method more generic in order to be used by page engine and others like blog,...

Could we do this  : 
```ruby
def title_with_translations(record)
  record.title.presence || record.translations.detect { |t| t.title.present?}.title
end
```

Instead of : 
https://github.com/refinery/refinerycms/blob/master/pages/app/helpers/refinery/admin/pages_helper.rb#L52
```ruby
def page_title_with_translations(page)
  page.title.presence || page.translations.detect { |t| t.title.present?}.title
end
```

And move it in another helper who is load on every admin controllers like : 
https://github.com/refinery/refinerycms/blob/master/core/app/helpers/refinery/admin_helper.rb

bricesanchez:
Fix #2870

### DIFF
--- a/core/app/helpers/refinery/translation_helper.rb
+++ b/core/app/helpers/refinery/translation_helper.rb
@@ -11,5 +11,8 @@ module Refinery
       val
     end
 
+    def translated_field(record, field)
+      Refinery::TranslatedFieldPresenter.new(record).call(field)
+    end
   end
 end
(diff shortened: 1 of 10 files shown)


parndt:
I like it, but do we need the helper method?

bricesanchez:
@parndt i like the helper method. IMHO it's simplier to understand than the presenter himself in the view.

PR #3128, opened 2016-03-08
Wait for action results before trying to use page controls.

parndt:
This fixes flaky tests on Travis CI as well as locally.
You can see the difference by running before/after this patch:

```
bundle exec rspec pages/spec/features/refinery/admin/pages_spec.rb -e "a page with two locales"
```


bricesanchez:
:+1:

PR #3117, opened 2016-02-22
Bugfix #3097 

bricesanchez:
It's ready to be reviewed and merged.

PR #3138, opened 2016-03-22
Add Rails::Html::PermitScrubber in section_presenter

bricesanchez:
This allow data-attributes in HTML

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,13 +3,20 @@
 * [See full list](https://github.com/refinery/refinerycms/compare/3-0-stable...master)
 
 
+## 3.0.3 [unreleased]
+
+* Fixed ability to display HTML data-attributes in content section presenter. [#3138](https://github.com/refinery/refinerycms/pull/3138). [Brice Sanchez](https://github.com/bricesanchez)
+
+* [See full list](https://github.com/refinery/refinerycms/compare/3.0.2...master)
+
+
 ## 3.0.2 [16 March 2016]
 
 * Update respond_to? and method_missing API. [21e77b08563905b1a79d335cfa3f32278961f24b](https://github.com/refinery/refinerycms/commit/21e77b08563905b1a79d335cfa3f32278961f24b). [Philip Arndt](https://github.com/parndt)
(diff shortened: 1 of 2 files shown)


anitagraham:
How does this interact with the `Refinery::Wymeditor.whitelist_tags`?

bricesanchez:
@anitagraham it does not interact because `Refinery::Wymeditor` is not mandatory on Refinery.

anitagraham:
Refinery::Wymeditor may not be mandatory, but that doesn't mean that there will be no interaction.

Whatever method I use to input HTML, if I use a tag not included by `Loofah:::HTML5::Whitelist::ALLOWED_ELEMENTS` who wins?
As Loofah is called at the rendering of the content it would appear that Loofah wins. 

So

* It may exclude elements which the user has whitelisted in Refinery::Wymeditor (or has added in some other way)
* the description of the PR should read "Limit elements to those permitted by Loofah:::HTML5::Whitelist::ALLOWED_ELEMENTS"
* Refinerycms-wymeditor should also note (somewhere) that the configuration is subject to the elements allowed by Loofah.
* Is there value in having a configuration, which, combined with the standard Loofah list will permit a site to use other elements or attributes?

I assume that the list of allowed_elements is fairly comprehensive so that - in general - there won't be a problem.

However, wymeditor has been the default editor for Refinery until now, so changes which will impact on commom usage should be noted.

*Note* I have passed information on using the `Wymeditor::whitelist_tags` to an inquirer on Gitter or Google Groups quite recently, so this is not necessarily an old or obsolete usage.

bricesanchez:
@anitagraham 

You're right, `Refinery::Wymeditor` should be updated to interact with `Loofah` or display a note because actually `Loofah` wins in all cases. 

This current PR fix this merged PR https://github.com/refinery/refinerycms/pull/3117 because data-attributes was not allowed.

parndt:
Great, this was causing major grief will trying to use recaptcha!

anitagraham:
https://github.com/flavorjones/loofah/issues/69 The source tag used inside the the `video` and `audio` tags is not supported by the Loofah Whitelist.
I am working on bringing `refinerycms-videojs` up to Refinery 3 compatibility and my source tags are being gutted. Not sure yet whether what I am seeing when editing the page is Loofah or Wymeditor, but it *will* be an issue when it comes to viewing the page.

PR #3145, opened 2016-03-30
Add page_parts factory

bricesanchez:
This should fix failling specs in https://travis-ci.org/refinery/refinerycms-page-images/builds/118346622

Capybara can not switch tabs because there is no page part.

Bonus : I've seen that Poltergeist now consider `opacity` for visibility  : https://github.com/teampoltergeist/poltergeist/pull/675


parndt:
@bricesanchez why are all pages required to have a page part now?  Shouldn't this be something that the developer specifies?

bricesanchez:
I've made some changes.

parndt:
Better.

Issue #3119, opened 2016-02-23
Issues with creating a custom action as described in guide


sturad:
I am having difficult in creating a custom action as described in 4.3 of creating a custom action  here  http://www.refinerycms.com/guides/using-custom-view-or-layout-templates.  I’m using Refinery 3.0.1and Rails 4.2.4, Ruby ruby-2.2.0. I create sample project just to retest this issue: [(https://github.com/sturad/refinery_custom_action_issue/)](https://github.com/sturad/refinery_custom_action_issue/)

Rails doesn't start with a  custom action defined by this routes.rb 

``` ruby
Refinery::Core::Engine.routes.prepend do
  get '/about-us', :to => 'pages#about_us', :as => :about_us
end
Rails.application.routes.draw do
  mount Refinery::Core::Engine, at: Refinery::Core.mounted_path
end
```

and this decorator #app/decorators/controllers/refinery/pages/pages_controller_decorator.rb

``` ruby
Refinery::PagesController.class_eval do
   skip_before_action :find_page, :only => [:about_us]
  def about_us
    # hardcoded id to debug query success
    @page = ::Refinery::Page.find(1)
    byebug
    render_with_templates?
  end
end
```

I get this error when starting Rails. 

```
/Users/sraden/.rvm/gems/ruby-2.2.0@picbase/gems/actionpack-4.2.4/lib/action_dispatch/routing/route_set.rb:557:in `add_route': Invalid route name, already in use: 'about_us'  (ArgumentError)
You may have defined two routes with the same name using the `:as` option, or you may be overriding a route already defined by a resource with the same naming. For the latter, you can restrict the routes created with `resources` as explained here: 
http://guides.rubyonrails.org/routing.html#restricting-the-routes-created
  from /Users/sraden/.rvm/gems/ruby-2.2.0@picbase/gems/actionpack-4.2.4/lib/action_dispatch/routing/mapper.rb:1561:in `add_route'
  from /Users/sraden/.rvm/gems/ruby-2.2.0@picbase/gems/actionpack-4.2.4/lib/action_dispatch/routing/mapper.rb:1536:in `decomposed_match'
  from /Users/sraden/.rvm/gems/ruby-2.2.0@picbase/gems/actionpack-4.2.4/lib/action_dispatch/routing/mapper.rb:1517:in `block in match'
[...]

PR #3146, opened 2016-03-30
Allow finders to be defined for each action.

parndt:
Per #3119 this will allow one to define a method such as:

```ruby
class PagesController
  module Finders
    class AboutUs
      def call(params)
        # logic for finding a page
      end
    end
  end
end
```

The advantage of this is that it allows any number of custom actions and somewhat removes the hardcoded nature of home and show.

Interested to see the results of CI for this change.

* [x] Needs tests

### DIFF
--- a/pages/app/controllers/refinery/pages_controller.rb
+++ b/pages/app/controllers/refinery/pages_controller.rb
@@ -68,12 +68,7 @@ module Refinery
     end
 
     def find_page(fallback_to_404 = true)
-      @page ||= case action_name
-                when "home"
-                  Refinery::Page.find_by(link_url: '/')
-                when "show"
-                  Refinery::Page.friendly.find_by_path_or_id(params[:path], params[:id])
-                end
+      @page ||= action_page_finder.call(params) if action_has_page_finder?
       @page || (error_404 if fallback_to_404)
     end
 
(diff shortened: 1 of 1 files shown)


pages/app/controllers/refinery/pages_controller.rb
```diff
@@ -3,7 +3,8 @@ class PagesController < ::ApplicationController
     include Pages::RenderOptions
 
     before_action :find_page, :set_canonical
-    before_action :error_404, :unless => :current_user_can_view_page?
+    before_action :error_404, if: :action_has_page_finder?,
```
gwagener: 404 _if_ there's a page finder?

parndt: This is a bit strange. I wonder how this is supposed to work :smile:

bricesanchez:
@parndt Are you able to test this feature for the milestone 3.0.4 ?

bricesanchez:
🆙 @parndt :)

parndt:
@bricesanchez it's on my list, but time is lacking.

parndt:
I'm actually a little puzzled about how to test this properly.

parndt:
Turns out the tests we have already adequately cover this

parndt:
@bricesanchez good to go

Issue #3148, opened 2016-04-06
Error if gem 'refinerycms' is added before 'bootstrap-sass' in the Gemfile


thebravoman:
In the Gemfile there is a line

    gem 'bootstrap-sass'

If you add

    gem 'refinerycms' 

before this line then the error occurs

If you add

    gem 'refinerycms' 

after this line then no error occurs.

The error is Sprockets::FileNotFound - couldn't find file 'bootstrap' with type 'application/javascript':

I have set up a test project for this
https://github.com/thebravoman/refinery_error_example

I have also created a stackoverflow question while investigating - http://stackoverflow.com/questions/36448308/refinery-sprocketsfilenotfound-couldnt-find-file-bootstrap-with-type-a/36462342#36462342

bricesanchez:
See https://github.com/thebravoman/refinery_error_example/pull/1 :)

thebravoman:
This might fix the problem. Also adding refinerycms after bootstrap-sass in the gem file does. 

But I could not find a single place where it is documented, mentioned, answered or just said that "bootstrap-sass" should be added to the application.rb.

anitagraham:
I think we should document that `gem refinerycms` and its friends should come at the end of the Gemfile.
Wording perhaps

In your Gemfile `'gem refinerycms'` should come after the standard inclusions, including Rails, database gems and asset gems.

Refinerycms extensions (such as `refinerycms-page-images` or `refinerycms-blog` should come after 'gem refinerycms`.

thebravoman:
Yes, that could help. But I am not sure and do not know the details  of why this is like this.

anitagraham:
Me neither... and I will try and find out.
When refinery installs it is always at the end, which is why most people don't come across this problem.

PR #3149, opened 2016-04-07
Note that refinery and associated gems should come after standard gems

anitagraham:
As discussed in reference to #3148


bricesanchez:
Thanks @anitagraham!

PR #3151, opened 2016-04-07
Add configs to handle SectionPresenter sanitizer

bricesanchez:
@anitagraham i've updated your work to add configs for SectionPresenter sanitizer.

@parndt and @anitagraham : Could you review it ?


anitagraham:
I have split the test so it tests elements and attributes separately, and minimised the HTML so it only includes the element/attribute being tested.

PR #3154, opened 2016-04-10
Html whitelist

anitagraham:
This is, I think, complete.
The tests work, and test only the add_whitelist features.
The code includes all of @bricesanchez changes.

If there are git knots I gnot up to untangling them. :smile:

### DIFF
--- a/pages/lib/refinery/pages/configuration.rb
+++ b/pages/lib/refinery/pages/configuration.rb
@@ -8,7 +8,8 @@ module Refinery
                     :cache_pages_backend, :cache_pages_full, :layout_template_whitelist,
                     :use_layout_templates, :page_title, :absolute_page_links, :types,
                     :auto_expand_admin_tree, :show_title_in_body,
-                    :friendly_id_reserved_words, :layout_templates_pattern, :view_templates_pattern
+                    :friendly_id_reserved_words, :layout_templates_pattern, :view_templates_pattern,
+                    :add_whitelist_elements, :add_whitelist_attributes, :whitelist_elements, :whitelist_attributes
 
     self.pages_per_dialog = 14
     self.pages_per_admin_index = 20
@@ -22,7 +23,19 @@ module Refinery
     self.cache_pages_backend = false
     self.cache_pages_full = false
     self.layout_template_whitelist = ["application"]
+    self.add_whitelist_elements = %w[ source track ]
+    self.add_whitelist_attributes = %w[ kind srclang placeholder controls ]
+
+
     class << self
+      def whitelist_elements
+        Loofah::HTML5::WhiteList::ALLOWED_ELEMENTS.merge(config.add_whitelist_elements)
+      end
+
+      def whitelist_attributes
+        Loofah::HTML5::WhiteList::ALLOWED_ATTRIBUTES.merge(config.add_whitelist_attributes)
+      end
+
       def layout_template_whitelist
         Array(config.layout_template_whitelist).map(&:to_s)
       end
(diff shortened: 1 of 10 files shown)


pages/lib/refinery/pages/configuration.rb
```diff
@@ -1,3 +1,5 @@
+require 'set'
```
parndt: where is `Set` used?

bricesanchez: I've forgotten to remove it, thanks for your review.

@anitagraham could you remove it and squash your commits ?

anitagraham:
I will make a couple more changes to the tests and squash.
I want to change the tests so that the elements and attributes being used will never be in the Loofah whitelist.

anitagraham:
Also I thought set was needed because the Loofah whitelist is a set. Will see what happens when I remove it.

anitagraham:
1 commit. 4 files. No 'require set'.

sintro:
What about cross-model solution (Refinery::Blog::Post, and probably user-defined extensions)?

bricesanchez:
@anitagraham Could rebase your commits ?

@sintro I'm not sure to understand your idea

sintro:
@bricesanchez, as I can assume from fast reading of commits, all the changes are focusing on Refinery::Page sanitizing when presenting it's content. Meanwhile Refinery::Blog::Post will be untouched and admin will not be able to set tags and attributes whitelists for Posts content, am I right?

bricesanchez:
@sintro The changes are focusing on the section presenter which is used to display blog post too : https://github.com/refinery/refinerycms-blog/blob/master/app/views/refinery/blog/posts/index.html.erb#L19

So `Refinery::Blog::Post` will be touched too and you will have to add missing elements/attributes in your Refinery::Page config.

parndt:
What if you just have the blog extension but not the pages extension? Sounds like coupling.

bricesanchez:
@parndt so there will be no sanitize

parndt:
My main concern is unrelated to this change which is that content_page relies on a helper in the pages extension, which is not guaranteed to be there 😦

bricesanchez:
You're right, in order to test coupling, we will probably need to extract all engines from the core?

parndt:
@bricesanchez you can do this now: 

```
rails _4.2.6_ new blog_test_app
```

If you put just the blog in your project's Gemfile:

```ruby
gem "refinerycms-blog", git: "https://github.com/refinery/refinerycms-blog"
```

This will merely pull in `refinerycms-core` and `refinerycms-settings` based on the current gemspec.  Then you can run:

```
bundle install
rails g refinery:cms
```

And all the other normal set up tasks like migrations, and you should have a Refinery with _only_ the Blog, and this _should_ fail when you open its index page, but also much before that:

```
undefined method `translates' for #<Class:0x007fd4e4b6de18>
```

😦

bricesanchez:
Is there a doc for this awesome thing?

parndt:
@bricesanchez no, it's just been quietly supported for a couple of major versions 😄

anitagraham:
@sintro If  you have a standard Refinery with Refinery-Blog extension, then the changes to the Pages config will also apply to the Blog posts as I think they both use the same code.  I will have to check that to be sure.

anitagraham:
@parndt, @bricesanchez 
You are right. The handling of HTML content/page parts should be separated from Pages. 

Pages:  Title, Page parts, control parts (now called advanced settings).

I would really like to discuss this more - as a route to making pages/parts/controls/extensions more othogonal to each other.

anitagraham:
PS. I don't understand the conflicts. If I follow the command line instructions nothing seems to change.

bricesanchez:
Closed in favor of #3164

PR #3156, opened 2016-04-11
Refinery::Core.google_analytics_page_code use ENV

bricesanchez:
That's a better practice to use a ENV variable instead of setting the Google Analytics code in the initializer.


parndt:
I don't agree, because this is not a secret token.  This gets placed publicly on every page.

Why is using an `ENV` better here? Which better practice is this following? 😄

bricesanchez:
I assume we does not need google analytics in local dev, i manage my UA on the server like the others ENV

PR #3157, opened 2016-04-11
Activate full-page-cache on production

bricesanchez:


parndt:
Are we saying that this will work in _every single production environment_?

bricesanchez:
It will work on production mode, i assume we does not need cache in dev mode.

parndt:
I'm :-1: on this.  It forces people to have a new dependency on `actionpack-page_caching` which they don't currently need.

bricesanchez:
Me too finally, i've forgot the dependency.

PR #3158, opened 2016-04-11
Set new_page_parts to true in Dev

bricesanchez:
It saves time to set `new_page_parts` to true in Dev only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 language: ruby
 cache: bundler
 bundler_args: --without development
-before_script: "bin/rake refinery:testing:dummy_app"
+before_script:
+  - export RETRY_COUNT=3
+  - bin/rake refinery:testing:dummy_app
 script:
-  - "bin/rspec $EXTENSION/spec"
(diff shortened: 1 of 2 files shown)


parndt:
Development mode should reflect production mode, so this creates a situation where a feature is expected to be available in production, because it's available in development mode, but then isn't. 

Doesn't this violate a principle of least surprise for our users?

bricesanchez:
My users never creates page parts, for me it's a developper feature. 

But if your template needs custom page parts, it will be better to implement the feature "page parts by template" instead of letting the user guess what is the good page part for his template.

bricesanchez:
i don't know what happens with this branch, i've rebased it and forced update, but i've made a mistake on  git, there is no merge on master.

parndt:
I'm still :-1: on this

parndt:
If the implementation were as you say, with templates specifying whether new page parts were allowed, then I'd support having this feature always enabled, without a config setting at all.

anitagraham:
Sorry @bricesanchez , I'm with @parndt on this. No surprises in production

PR #3161, opened 2016-04-14
Add doc standalone extension in your Rails Application

bricesanchez:


parndt:
@bricesanchez great stuff! I don't think we can merge it until that process actually works without errors.

bricesanchez:
I will try take time to fix this process :)

parndt:
also I think using `[ci skip]` means the required build pass step never finishes

bricesanchez:
Should we avoid using `[ci skip]` or could you disable required CI when we use `[ci skip]` ?

parndt:
I looked into it, but we can't get that granular.  I suppose, at that point, it'd be nice if Travis just reported success if `[ci skip]` were enabled.

PR #3164, opened 2016-04-19
Html whitelist

bricesanchez:
This PR replaces #3154, it's just a rebase on master.

### DIFF
--- a/pages/app/presenters/refinery/pages/section_presenter.rb
+++ b/pages/app/presenters/refinery/pages/section_presenter.rb
@@ -64,9 +64,8 @@ module Refinery
 
       def wrap_content_in_tag(content)
         content = sanitize(content,
-          tags: Loofah::HTML5::WhiteList::ALLOWED_ELEMENTS,
-          attributes: Loofah::HTML5::WhiteList::ALLOWED_ATTRIBUTES,
-          scrubber: Rails::Html::PermitScrubber.new
+          tags: Refinery::Pages::whitelist_elements,
+          attributes: Refinery::Pages::whitelist_attributes
         )
         content_tag(:section, content_tag(:div, content, :class => 'inner'), :id => id)
       end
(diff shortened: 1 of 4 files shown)


anitagraham:
Thankyou for the rescue @bricesanchez

bricesanchez:
You're welcome!

bricesanchez:
@parndt, it's ready to be merged if you're ok.

sintro:
Ok, and how I can remove some elements from whitelist? If I want to disable the possibility of including <script> in pages body?

parndt:
`<script>` tags should already be disabled?

anitagraham:
Well I suppose, if it were required we could include a blacklist if there was a demand (taking @parndt 's comment into consideration).

sintro:
@parndt , I just tried to save ``<script>alert('hello')</script>;`` (by switching to HTML editor), and after that I successfully alerted when opened that page.
By reading this https://github.com/refinery/refinerycms/blob/c793da02180489f44f491804fe61ebfd7c127f58/pages/lib/refinery/pages/configuration.rb#L32 line I see, that there are only  Loofah::HTML5::WhiteList::ALLOWED_ELEMENTS whitelisted elements, which does not include 'script' tag, but probably it is rendered by ``=raw @page.content_for(:body)``, which I use in page\`s template.

bricesanchez:
@sintro 

> Loofah::HTML5::WhiteList::ALLOWED_ELEMENTS whitelisted elements, which does not include 'script' 
Which is not explicitly allowed is forbidden.

> So I am sure it is rendered by =raw @page.content_for(:body)
The HTML sanitizer is for frontend only, so <script> tags could be recorded in backend but they will be sanitize in frontend.

sintro:
@bricesanchez , exactly, that is what I have noticed in https://github.com/refinery/refinerycms/pull/3164#issuecomment-226680890, but the question is still here: what if I want to disable some of ALLOWED_ELEMENTS`s default tags?

bricesanchez:
@sintro Ok, i understand. So you will have to override the `Refinery::Pages` configurable class if you want to disable more tags in frontend :

https://github.com/refinery/refinerycms/blob/c793da02180489f44f491804fe61ebfd7c127f58/pages/lib/refinery/pages/configuration.rb#L31-L33

ttmthanh:
@bricesanchez How can I override it? I must create new modul and copy all about file configuration or something else? Please help me :'(

PR #3168, opened 2016-05-01
Add preview skip_to_first_child page message

bricesanchez:


pages/config/locales/en.yml
```diff
@@ -40,6 +40,8 @@ en:
           view_live_html: View this page live <br/><em>(opens in a new window)</em>
           hidden: hidden
           draft: draft
+        preview:
+          skip_to_first_child: "The page %{what} use the option \"Skip to first child\" and will never be displayed to guest users."
```
parndt: `uses` not `use` here
`guest users` should be `visitors`

core/app/views/layouts/application.html.erb
```diff
@@ -12,6 +12,9 @@
         <%= render "/refinery/header" -%>
       </header>
       <section id="page">
+        <div id="flash_container">
+          <%= render '/refinery/message' %>
+        </div>
```
parndt: Unfortunately, I am very against adding the flash container onto the application layout.  I feel that this may lead to flash notices bleeding through to users. Is it possible to only have this for the preview controller?

core/app/assets/javascripts/refinery/core.js.erb
```diff
@@ -1,11 +1,12 @@
 $(document).ready(function(){
   // focus first field in an admin form.
   $('form input[type=text]:first').focus();
-
-  init_flash_messages();
```
parndt: what invokes this now?

bricesanchez: I've seen that it was call 2 times before i remove this call, the first call was : https://github.com/refinery/refinerycms/blob/c28e7afae816328280fb6f511094d8cbf100ebd0/core/app/assets/javascripts/refinery/admin.js.erb#L6

core/app/views/layouts/application.html.erb
bricesanchez: Let me check that

core/app/assets/javascripts/refinery/core.js.erb
parndt: So it'll only be run inside layouts including `admin.js` ? Doesn't that mean _not_ the Preview controller?

pages/config/locales/en.yml
bricesanchez: Fixed and will be push.

core/app/assets/javascripts/refinery/core.js.erb
bricesanchez: Yes and i think preview_controller needs refactoring : 

* Missing `:custom_slug` param : https://github.com/refinery/refinerycms/blob/c28e7afae816328280fb6f511094d8cbf100ebd0/pages/app/controllers/refinery/pages/admin/preview_controller.rb#L40-L46

* Hardcoded layout instead of page.layout with fallback : https://github.com/refinery/refinerycms/blob/c28e7afae816328280fb6f511094d8cbf100ebd0/pages/app/controllers/refinery/pages/admin/preview_controller.rb#L36-L38

parndt: can it be done as part of this PR?

bricesanchez: Are you available on Gitter?

bricesanchez:
Ready to be reviwed (with tests)

parndt:
Thanks! I know this might be annoying, but can you add a test (here: https://github.com/refinery/refinerycms/blob/master/pages/spec/features/refinery/pages_spec.rb#L320-L338) for when you visit the page normally (as a visitor) to make sure that we _don't_ see the message (i.e. that the page that's redirected to doesn't somehow display it?)

Mainly to protect against regressions.

bricesanchez:
No problem, it's done.

bricesanchez:
I don't understand, none of my tests fails on local, is there a problem with the Travis config ?

parndt:
Very rarely is Travis the problem.. I'll look at it in a few hours

parndt:
The problem was Bundler versions 1.12.0 through 1.12.3.  Fixed in 1.12.4

bricesanchez:
Thanks @parndt for find the solution!

PR #3180, opened 2016-05-17
Ensure the dummy app uses the same template as real apps.

parndt:
**This Fails**

This is work I was doing last July, but it still seems like a good thing to work on.  I wasn't happy that the dummy app used a different template than the one we ask people to install with because this way we can implicitly test that template's always correct.

It needs rebasing and validating.

### DIFF
--- a/core/lib/refinery/core/environment_checker.rb
+++ b/core/lib/refinery/core/environment_checker.rb
@@ -1,0 +1,54 @@
+# Check that Rails application configuration is consistent with what we expect.
+require 'pathname'
+
+module Refinery
+  module Core
+    class EnvironmentChecker
+
+      def initialize(root_path, environments: %w(development test production))
+        @root_path = Pathname.new(root_path)
+        @environments = environments
+      end
+
+      def call
+        environment_files.each do |env_file|
+          # Refinery does not necessarily expect action_mailer to be available as
+          # we may not always require it (currently only the authentication extension).
+          # Rails, however, will optimistically place config entries for action_mailer.
+          current_mailer_config = mailer_config(env_file)
+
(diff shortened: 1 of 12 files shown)


core/lib/refinery/core/environment_checker.rb
```diff
@@ -0,0 +1,54 @@
...
+        @root_path = Pathname.new(root_path)
+        @environments = environments
+      end
+
+      def call
+        environment_files.each do |env_file|
+          # Refinery does not necessarily expect action_mailer to be available as
+          # we may not always require it (currently only the authentication extension).
+          # Rails, however, will optimistically place config entries for action_mailer.
+          current_mailer_config = mailer_config(env_file)
```
parndt: This mailer stuff should be done by refinerycms-inquiries. The core doesn't send mail. 

Maybe this whole thing shouldn't even be our concern?

core/lib/refinery/core/environment_checker.rb
```diff
@@ -0,0 +1,54 @@
...
+              "  end"
+            ].join("\n")
+
+            env_file.write(
+              env_file.read.gsub(current_mailer_config, new_mailer_config)
+            )
+          end
+
+          env_file.write(
+            env_file.read.gsub("assets.compile = false", "assets.compile = true")
```
parndt: We don't neee this

bricesanchez:
Hi @parndt ! Could you update your PR?

parndt:
@bricesanchez please!

bricesanchez:
I've tried to fix specs on this PR but i'm not sure if my commits helps. Please, could you check if you can fix specs @parndt ?

PR #3183, opened 2016-05-23
Allow regex in attribute whitelist.

nzgrover:
Adds the ability to use a regex in the attribute whitelist. The specific need I have is to allow all 'data-whatever' attributes through. 
Just wanted to get the ball rolling with this commit, look forward feedback on improvements or the 'refinerycms' way to do this.


parndt:
Thanks @nzgrover 

Is there no support in Loofah to do this?

bricesanchez:
@nzgrover Could rebase your branch with the latest commit on master ?

Thanks!

nzgrover:
@parndt I couldn't find anything in loofah's documentation, and there were data-* attributes being removed, but....
https://github.com/flavorjones/loofah/pull/84
https://github.com/flavorjones/loofah/blob/master/CHANGELOG.rdoc
Says they are supported in version 2.0.2
I'll investigate further,

parndt:
@bricesanchez no point rebasing until it's passed review 😄 
@nzgrover thanks a lot!

bricesanchez:
@parndt my mistakes i was searching something in the wrong file. I thought a rebase will display it..

core/lib/refinery/core/configuration.rb
```diff
@@ -32,6 +32,7 @@ module Core
     self.visual_editor_javascripts = []
     self.visual_editor_stylesheets = []
     self.plugin_priority = []
+    self.regex_white_list = false
```
bricesanchez: All configs for SectionPresenter are in the `pages/lib/refinery/pages.configuration`. I'm in favor to move `regex_white_list ` in Pages and rename it `whitelist_regex` to match current configs.

Could you add by default your regex?

parndt: I still want to know whether we can support this in Loofah with existing config options

pages/app/presenters/refinery/pages/section_presenter.rb
```diff
@@ -69,10 +69,15 @@ def wrap_content_in_tag(content)
       end
 
       def sanitize_content(input)
-        output = sanitize(input,
-          tags: Refinery::Pages::whitelist_elements,
-          attributes: Refinery::Pages::whitelist_attributes
-        )
+        output =
+            if Refinery::Core.regex_white_list
```
bricesanchez: For me `whitelist_regex` should be a config and not a boolean.

No need to make and if else statement, we should keep the CustomScrubber only.

pages/app/presenters/refinery/pages/section_presenter.rb
```diff
@@ -85,5 +90,27 @@ def sanitize_content(input)
         return output
       end
     end
+
+    class CustomScrubber < Rails::Html::PermitScrubber
```
bricesanchez: Is it the right place for this class ? @parndt

parndt: Sure, this placement is fine, or it could go in its own file.

core/lib/refinery/core/configuration.rb
bricesanchez: I think it still use `Loofah` for the configurations `Refinery::Pages::whitelist_elements`and `Refinery::Pages::whitelist_attributes` but use `Rails::Html::PermitScrubber` to create the scrubber intead of `Loofah::Scrubber`

nzgrover:
@parndt I've had a further look at seeing if we can just use loofah, but as far as I can tell, if you supply attributes then ONLY those attributes explicitly supplied are allowed through. If you don't supply attributes, then the loofah default scrubber is used, which includes a regex to allow all 'data-' attributes.
The only way I can think of round it would be to update loofahs default allowed attributes directly, but loofah doesn't provide a way to do so. See this very old issue open regarding this: https://github.com/flavorjones/loofah/issues/14
How do you wish to proceed?

nzgrover:
Or what about something (slightly cheeky) like this: https://github.com/refinery/refinerycms/compare/master...nzgrover:allow_all_data_attributes

parndt:
@nzgrover I really like that solution.. allowing data- attributes seems sane; we'll need to document this next to the existing config options I think something like:

```ruby
# Note: data- attributes are whitelisted by default. See (link to PR or something)
config.whitelist_attributes = []
```

nzgrover:
@parndt Ok, shall I put in a new pull request with the allow_all_data_attributes branch?

parndt:
@nzgrover yes please

PR #3185, opened 2016-05-24
Bugfix/3184 : Fix rubydocs

bricesanchez:
@parndt Could you use this branch to deploy rubydocs and check if it fix the problem?

c.c. @kholbekj


parndt:
I'm not sure what is required to make it start working now.. it should work.

kholbekj:
Yeah now it's just being troublesome... You're also seing the "invalid git url" thingie right?

parndt:
Yes! Perhaps it's not allowed after silently failing for so long.. I don't know.

@lsegal sorry to interrupt you Loren, but do you happen to know why we can't seem to add this project to rubydoc.info ? ❤️ thanks for your time

PR #3187, opened 2016-05-26
Allow all data attributes

nzgrover:
@parndt New PR to allow all "data-" attributes.


pages/app/presenters/refinery/pages/section_presenter.rb
```diff
@@ -85,5 +83,23 @@ def sanitize_content(input)
...
+      def initialize(tags, attributes)
+        @direction = :bottom_up
+        @tags = tags
+        @attributes = attributes
+      end
+
+      #see https://github.com/flavorjones/loofah/blob/master/lib/loofah/html5/scrub.rb#L21
+      def scrub_attributes(node)
+        node.attribute_nodes.each do |attr_node|
+          next if attr_node.node_name =~ /\Adata-[\w-]+\z/
```
bricesanchez: `/\Adata-[\w-]+\z/` Why do we hardcoded this regex and don't let the dev to be able to override it in a config?

pages/lib/refinery/pages/configuration.rb
```diff
@@ -23,6 +23,7 @@ module Pages
     self.cache_pages_full = false
     self.layout_template_whitelist = ["application"]
     self.add_whitelist_elements = %w[ source track ]
+    # Note: "data-" attributes are whitelisted by default. See ...
```
bricesanchez: It missing the `See ...` link

nzgrover:
@bricesanchez @parndt suggested he preferred this approach and asked me to submit a new pull request.

parndt:
@nzgrover yes, I think what @bricesanchez is asking is whether we can move the regular expression itself to be configurable.  I don't personally think we need to worry about that as it's easy enough for someone to write their own Presenter and Scrubber.  Thanks for your work!

sintro:
Hmm, looks like "render_content_page" method from '/refinery/content_page' template still removes the data-attributes. 
Here is the fragment of log warning after sanitizing:
```
 <div class="image">
-        <img width="300" height="65" alt="Slide Background 2" title="Р•СЃС‚СЊ Р·Р°РіРѕР»РѕРІРѕРє" data-no-retina="true" src="/system/images/W1siZiIsIjIwMTYvMDQvMDgvMmp5dHZuaHJpZl9zbGlkZV9iYWNrZ3JvdW5kXzQuanBnIl0sWyJwIiwidGh1bWIiLCIzMDB4MjAwIl1d/slide-background-4.jpg?sha=b4c8abc2dfc3f003" />
+        <img width="300" height="65" alt="Slide Background 2" title="Р•СЃС‚СЊ Р·Р°РіРѕР»РѕРІРѕРє" src="/system/images/W1siZiIsIjIwMTYvMDQvMDgvMmp5dHZuaHJpZl9zbGlkZV9iYWNrZ3JvdW5kXzQuanBnIl0sWyJwIiwidGh1bWIiLCIzMDB4MjAwIl1d/slide-background-4.jpg?sha=b4c8abc2dfc3f003">
```
As one can see here, Scrubber has removed the 'data-no-retina' attruibute. 

PS
I do have lines
```
def scrub_attributes(node)
        node.attribute_nodes.each do |attr_node|
          next if attr_node.node_name =~ /\Adata-[\w-]+\z/

          super
        end
      end
```
in my section_presenter.rb

PR #3200, opened 2016-06-17
Assets for glass admin index views

stefanspicer:
Supporting css & partials for glass admin index views.  The first one is here: https://github.com/refinery/refinerycms-authentication-devise/pull/23

### DIFF
--- a/core/app/assets/stylesheets/glass/layout/_pages.scss
+++ b/core/app/assets/stylesheets/glass/layout/_pages.scss
@@ -1,0 +1,508 @@
+/* -------------------- COMMON -------------------- */
+
+.container-fluid {
+  @include media-breakpoint-up(sm) {
+    padding-left: 60px;
+  }
+}
+
+.admin-title {
+  margin: 15px 0 30px;
+  font-weight: 600;
+  text-align: center;
+  letter-spacing: 2px;
+  text-transform: uppercase;
+}
+
+.contextual-menu {
+  display: flex;
+  align-items: center;
+  margin-top: 15px;
+  margin-bottom: 30px;
+  padding: 7px 15px;
+  border-radius: $radius;
+  background: $white;
+  & > * {
+    margin-top: 8px;
+    margin-bottom: 8px;
+  }
+  .contextual-options {
+    display: flex;
+    align-items: center;
+    width: 100%;
+  }
+  .relative-nav-item {
+    min-width: 80px;
+  }
+  .btn-group .btn-link-gray {
+    border: 1px solid $gray-3;
+  }
+  .btn-group {
+    .btn {
+      margin: 0;
+    }
+  }
+  .btn {
+    @include transition-fix();
+    &:first-child {
+      margin-left: 0;
+    }
+    &:last-child {
+      margin-right: 0;
+    }
+  }
+  .btn-link-gray {
+    &:first-child {
+      margin-right: 0;
+      margin-left: 0;
+    }
+    &:last-child {
+      margin-right: 0;
+      margin-left: 0;
+    }
+    .icon {
+      color: $gray-3;
+    }
+  }
+  .btn {
+    margin-right: 7px;
+    margin-left: 7px;
+  }
+  & > .btn,
+  & > .dropdown {
+    &:last-child {
+      margin-right: 0;
+      margin-left: auto;
+      @media (max-width: 767px) {
+        order: -1;
+        margin-left: 0;
+      }
+    }
+  }
+  .dropdown {
+    margin: 0 7px;
+    &:hover,
+    &:focus,
+    &.open {
+      z-index: 1000;
+      outline: none;
+      .btn {
+        z-index: 1005;
+      }
+      .dropdown-menu {
+        visibility: visible;
+        height: auto;
+        transition: opacity .2s ease-in;
+        opacity: 1;
+      }
+    }
+    &:last-child {
+      margin-right: 0;
+      margin-left: auto;
+      .dropdown-menu {
+        right: 0;
+        left: auto;
+      }
+    }
+  }
+  .dropdown-menu {
+    z-index: 50;
+    top: -2px;
+    display: block;
+    visibility: hidden;
+    overflow: hidden;
+    min-width: 100%;
+    height: 0;
+    margin-bottom: 15px;
+    padding-top: 40px;
+    opacity: 0;
+    border-radius: 0;
+    a {
+      display: block;
+      padding: 5px 12px;
+      font-size: 14px;
+      transition: 0;
+      &:first-child {
+        border-top: 1px solid $gray-1;
+        &:hover {
+          border-color: transparent;
+        }
+      }
+      &:hover,
+      &:focus {
+        color: $white;
+        background: $cms-hover;
+        .icon {
+          color: $white;
+        }
+      }
+    }
+    .icon {
+      color: $gray-3;
+    }
+  }
+  .open {
+    .btn-link-gray {
+      z-index: 60;
+    }
+  }
+}
+
+@media (max-width: 767px) {
+  .contextual-menu {
+    flex-wrap: wrap;
+    justify-content: space-around;
+  }
+}
+
+.modal {
+  .form-actions,
+  .deliver,
+  .image-delete-btn {
+    /* Actions are handled by modal buttons.. remove the default ones */
+    display: none;
+  }
+}
+
+
+
+/* -------------------- PAGES -------------------- */
+
+// ADMIN PAGES
+
+.admin-page {
+  overflow: hidden;
+  padding-top: 60px;
+}
+
+.admin-profile-form {
+  .admin-title {
+    display: none;
+  }
+}
+
+
+// ADMIN INDEX
+
+.admin-index {
+  padding-top: 60px;
+  & > .row {
+    min-height: 450px; /* need room for dropdowns */
+  }
+  .index-listing {
+    margin-bottom: 90px;
+    @media (min-width: 768px) {
+      margin-bottom: 60px;
+    }
+    & > ul {
+      overflow: hidden;
+      border-radius: $radius;
+      li:last-child {
+        margin-bottom: -2px;
+      }
+    }
+  }
+  ul {
+    li {
+      margin-bottom: 0;
+      &:last-of-type {
+        padding-bottom: 0;
+      }
+      .page-row {
+        display: flex;
+        align-items: center;
+        margin-bottom: 2px;
+        padding: 0 15px;
+        font-size: 18px;
+        color: $gray-6;
+        background: $white;
+        &:hover {
+          background: darken($off-white, 1%);
+        }
+        @media (min-width: 768px) {
+          &:hover {
+            background: $cms-hover;
+            .icon {
+              color: $white;
+            }
+            .actions {
+              display: block;
+            }
+            .page-title {
+              color: $white;
+              .label {
+                color: $cms-hover !important;
+                background: $white !important;
+              }
+            }
+            .date {
+              display: none;
+            }
+            a.attribute {
+              color: $white;
+              &:hover {
+                text-decoration: underline;
+              }
+            }
+          }
+        }
+        &.no-edit {
+          .actions {
+            display: block;
+            .circle-icon {
+              border-color: $cms-color;
+              &:link {
+                color: $cms-color;
+              }
+              &:hover {
+                color: $white;
+                .icon {
+                  color: $white;
+                }
+              }
+              .icon {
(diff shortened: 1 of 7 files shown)


stefanspicer:
I just realized that the refinery/admin/_search.html.erb is probably going to make other views using that partial look weird, but I didn't look at them.  I should probably create a new temporary glass/admin/_search that can then be moved into acts-as-indexed when all search boxes are ported over to the new layout.  Good plan?

core/app/assets/stylesheets/glass/layout/_pages.scss
```diff
@@ -0,0 +1,647 @@
...
+    .btn {
+      margin: 0;
+    }
+  }
+  .search {
+    z-index: 100;
+    flex-order: 2;
+    width: 100%;
+    height: 40px;
+    @media (min-width: 768px) {
```
bricesanchez: Can't we use a variable instead of defining the width each time ?

core/app/assets/stylesheets/glass/layout/_pages.scss
```diff
@@ -0,0 +1,647 @@
...
+    margin-bottom: 2px;
+    padding: 10px 0 0 15px;
+    border: 1px dashed #cccccc;
+    background: lighten($off-white, 4%);
+    &:before {
+      position: relative;
+      top: 2px;
+      padding-left: 0;
+      font-family: "Refinery" !important;
+      font-size: 18px;
```
bricesanchez: I'm strongly in favor of `em` instead of `px` units and mostly on `font-size` : https://www.w3.org/TR/WCAG20-TECHS/C14.html

core/app/assets/stylesheets/glass/layout/_pages.scss
```diff
@@ -0,0 +1,647 @@
...
+      color: $gray-6;
+
+      -webkit-font-smoothing: antialiased;
+      -moz-osx-font-smoothing: grayscale;
+      speak: none;
+    }
+    &:after {
+      padding-left: 10px;
+      font-size: 18px;
+      content: "Move Here";
```
bricesanchez: Hardcoded english text will be a problem for others languages.

core/app/views/refinery/admin/_search.html.erb
```diff
@@ -1 +1,17 @@
-<%= render 'refinery/acts_as_indexed/admin/search', local_assigns if defined?(Refinery::ActsAsIndexed::Engine) -%>
```
bricesanchez: I think we should keep the partial as it and create a PR in https://github.com/refinery/refinerycms-acts-as-indexed/blob/master/app/views/refinery/acts_as_indexed/admin/_search.html.erb

bricesanchez:
Nice job, it will be easy to merged after the Q&A changes :)

core/app/views/refinery/admin/_search.html.erb
stefanspicer: Note that this currently (and an acts-as-indexed PR would as well) breaks other search forms.

refinery-search-wierd

And my next PR would break them even more.  So I've pushed what I suggest do to handle this: use a _tmp_search until all views have been moved over to use this new form.  What do you think of that?

A side note: It seems odd that the html for this lives in refinerycms-acts-as-indexed.  Is it just me, or does it make sense for refinerycms-acts-as-indexed to own the logic of searching and not the UI?

core/app/views/refinery/admin/_tmp_search.html.erb
```diff
@@ -0,0 +1,17 @@
...
+      <%= hidden_field :dialog, nil, :name => 'dialog', :id => nil, :value => true if from_dialog? %>
+      <%= hidden_field :callback, nil, :name => 'callback', :id => nil, :value => @callback if @callback.presence %>
+      <span id="refinery-search-btn" class="search-btn">
+        <button type="submit"><i class="icon icon-search"></i>
+          <div class="inline loader xs bump-up"></div>
+        </button>
+      </span>
+    </div><!-- /input-group -->
+  </form>
+</div>
```
stefanspicer: I think this comment was deleted (when the line diff was removed) so re-posting...

Note that my original PR (as well as your suggestion of an acts-as-indexed PR) breaks other search forms.

refinery-search-wierd


And my next PR would break them even more. So I've pushed what I suggest do to handle this: use a _tmp_search until all views have been moved over to use this new form. What do you think of that?

A side note: It seems odd that the html for this lives in refinerycms-acts-as-indexed. Is it just me, or does it make sense for refinerycms-acts-as-indexed to own the logic of searching and not the UI?

core/app/assets/stylesheets/glass/layout/_pages.scss
lukeschenk: I'll work on switching them to rem's. The relativity to the root font size makes things a lot more simple.

core/app/assets/stylesheets/glass/layout/_pages.scss
bricesanchez: friendly 🆙

core/app/views/refinery/admin/_tmp_search.html.erb
bricesanchez: I'm not in favor of a temporary partial name like `_tmp_search.html.erb` and personaly i use Elasticsearch in frontend and probably soon in backend. So i don't know if this partial should be in core or in an extension.

Any thoughts @parndt ?

parndt: Probably https://github.com/refinery/refinerycms-search should own it

stefanspicer: I'm fine to move this somewhere else.  Wherever it goes, there is going to be a temporary conditional logic to show the current search form html, and the new search form html.  You'd like to see that within refinerycms-search?  I don't see refinerycms-search in my Gemfile.lock.  Would we then be adding this as a new dependency?

parndt: Nope, I was proposing that we disable search unless the extension is added.

I'm trying to slim the core down as much as possible (it is still too bloated), and this seems like a good opportunity.

stefanspicer: I see.  I'll take a crack at it.

core/app/assets/stylesheets/glass/layout/_pages.scss
```diff
@@ -0,0 +1,539 @@
...
+        .page-row {
+          background: $white;
+        }
+        &:last-child {
+          .page-row {
+            margin-bottom: 4px;
+          }
+        }
+      }
+      li .page-row {
```
bricesanchez: I'm not sure about this part, it looks like the nested `li .page-row` level is hardcoded. What happens after 6 levels ?

bricesanchez:
Only one concern about nested li's, otherwise it's ok for me.

core/app/assets/stylesheets/glass/layout/_pages.scss
stefanspicer: I agree the nested li's hardcoded to 6 levels is bad.  I remember attempting to get rid of that a while back (just looked, it was late 2014 so it's been a while).  I ran into an issue at that point and couldn't get it to work without this.  Good news is users can't be nested (this is only for a tree like pages) so shall I just remove this until we get to the pages index?  At that point, please remind me to address this if I haven't already.

bricesanchez:
@stefanspicer thanks for the changes, could you rebase your branch and squash your commits?

stefanspicer:
How's that?  Good to go?  Do you prefer to have separate commits as it is in progress so the commits match the comments and conversation (and then squash at the end when you say to)?  Or shall I just squash everything as I have updates?

parndt:
Please don't squash as you go 😄 it's helpful to see the progress of changes.  It's also really useful for bisecting when we have little atomic commits that only each change one thing or introduce a single feature.

core/app/views/layouts/refinery/admin.html.erb
```diff
@@ -2,22 +2,30 @@
...
     <div id='tooltip_container'></div>
-    <div id="admin_container" class="clearfix">
-      <div id="page_container">
-        <div id="page">
-          <div id="content" class="clearfix">
-            <div id="flash_container">
-              <%= render '/refinery/no_script' %>
-              <%= render '/refinery/message' %>
+    <section id="wrapper" class="dimmer">
+      <% if content_for?(:admin_content) %>
```
parndt: Basically, what's the case for this new feature and how is it tested? :smile: Where is this used? Is this documented somewhere?

This also appears to remove the standard `yield` - does overriding `admin_content` mean you have to do that yourself?

stefanspicer: This feature, as I said in the initial comment is "Supporting css & partials for glass admin index views. The first one is here: refinery/refinerycms-authentication-devise#23".  If you look at that other PR, I have a screenshot that shows the new look.  I don't like the idea of any test for styles, so if all current tests pass without regression I think that's all we can expect for this one.

I removed the standard `yield` as a simple way to have old and new layouts working at the same time and I simply don't return anything with the standard yeild whenever I use the `admin_content`.  If there is a better way for a view to change a layout condition other than `if content_for?` we can use that.  I thought this code read well.

It may help to see where I'm going with this too.  Here's what's to come if we get here.  This includes 3 sections in the admin interface: A right sidebar (somewhat like the left sidebar with the nav), a page preview (this would yield a frontend `show` view to edit within), and then `admin_content` (index views, forms that don't make use of the preview).

```
    <% if content_for?(:admin_sidebar) %>
      <section id="sidebar-right" class="sidebar right off-canvas-sidebar cms-ui">
        <div class="sidebar-top-nav">
           <span>Publish settings</span>
           <button type="button" class="btn sidebar-right-close"><i class="icon icon-close"></i></button>
        </div>
        <div id="sidebar-right-inner" class="sidebar-inner">
          <%= yield :admin_sidebar %>
        </div>
      </section>
    <% end %>

    <section id="wrapper" class="dimmer dimmed-editing">
      <% if content_for?(:admin_page_preview) %>
        <section id="page-preview">
          <div class="page-container">
            <%= yield :admin_page_preview %>
          </div>
        </section>
      <% elsif content_for?(:admin_content) %>
        <div class="cms-ui">
          <%= yield :admin_content %>
        </div>
      <% end %>
    </section>
```

parndt: Thanks!

bricesanchez: > How's that? Good to go? Do you prefer to have separate commits as it is in progress so the commits match the comments and conversation (and then squash at the end when you say to)? Or shall I just squash everything as I have updates?

I've spoke about to clean commits log of the PR. Not squashing all commits into one. Sorry for the misunderstood

For example, if the first commit add features 1-2-3 and the second commit remove the feature 3, you can squash the commit to have only 1 commit with features 1-2.

stefanspicer: Here's what I'm thinking

```
pick 0dc3d38 Allow for admin index views
s b80ebdc A temporary search partial --> a modification of 0dc3d38
s 643d583 media queries, rems over px's --> a modification of 0dc3d38
s 9c4db16 rm tmp_search partial. Coming soon in refinerycms-search --> revert b80ebdc
s 5fbaeef rm search styles as they belong in refinerycms-search --> a modification of 0dc3d38
s 126bca9 a nice light gray background --> forgot to include this in 0dc3d38
s 416ed28 rm styles for index with nested rows (a tree) --> a modification of 0dc3d38
```

Which of these would you like to see as its own commit?  I think 643d583 is the only possibility, but it is really just a modification of 0dc3d38.  126bca9 is sort of a feature, but I always intended it to be in the first commit.

bricesanchez:
```
pick 0dc3d38 Allow for admin index views
s 416ed28 rm styles for index with nested rows (a tree)
s 5fbaeef rm search styles as they belong in refinerycms-search
pick 643d583 media queries, rems over px's
pick 126bca9 a nice light gray background
```

Probably more like this.

stefanspicer: I can do that.  In that case it seems you'd like to have 0dc3d38 split into 7-10 different commits more like 126bca9?

stefanspicer: I squashed that this morning as requested.  That looks better now?

bricesanchez: For me yes and you @parndt ?

parndt: Seems fine

PR #3203, opened 2016-06-29
Improve execjs error message

bricesanchez:
Beginners on Refinery CMS often forget to install `execjs`gem in order to properly create their application.

This PR improves the ERROR message and should reduce the number of questions on gitter and github.

refinery-without-execjs_ _bricesanchez_macbook-pro-de-alexandra_ _zsh_ _158x26

### DIFF
--- a/templates/refinery/demo.rb
+++ b/templates/refinery/demo.rb
@@ -5,7 +5,9 @@ begin
   require 'execjs'
   ::ExecJS::RuntimeUnavailable
 rescue LoadError
-  abort "ExecJS is not installed. Please re-start the installer after running:\ngem install execjs"
+  abort <<-ERROR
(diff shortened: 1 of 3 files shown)


parndt:
This is good, I wish we could DRY this between the templates but unfortunately we can't share code between them!

stefanspicer:
I've always used therubyracer.  I thought I got that from a guide somewhere.  Within refinery, is there a preference between execjs or therubyracer?

parndt:
We use execjs to detect the JS runtimes available. If you [look at the template file](https://github.com/refinery/refinerycms/blob/b97e84e0fb3f7e40db7fea04c98847e1e2d9afe5/templates/refinery/edge.rb#L1-L13), you'll see it's trying to put `therubyracer` in the Gemfile 😄

PR #3209, opened 2016-07-05
Whitelist select attribute: multiple

### LINKED ISSUE #3206: Select tag with multiple option failed
Hi there, 
 
I have select tag in the content_for Body,

When i add multiple attribute into select tag and save the page, "multiple" is disappeared, how can i add multiple attrubute?

bricesanchez:
It should fix #3206.


anitagraham:
Hmmm... I would be surprised if `multiple` was not on the default whitelist from Loofah.
[This](http://www.rubydoc.info/github/flavorjones/loofah/Loofah/HTML5/WhiteList) shows it in the attribute whitelist.

@AlbaHoo Does your log file have a message showing sanitization has taken place?
(Looks like this: `"HTML attributes and/or elements content has been sanitized`)

bricesanchez:
@anitagraham You are right, i did not check the doc before creating this PR. I close it

PR #3211, opened 2016-07-08
flash notices for glass index views

stefanspicer:
I left out the flash notices in #3200 by accident.  This adds them in.


bricesanchez:
Thanks @stefanspicer!

PR #3212, opened 2016-07-08
ExtensionHooks: more flexible & granular customization of refinery

stefanspicer:
I've been using a version of this for 2 years, and it has worked well.  I just refactored and improved upon it today for this PR.  With something like this, I can see us shrinking refinerycms-core as it allows us to easily move things into different extensions (like refinerycms-search).
## Custom js & css

Extensions can register js & css to be compiled in with:

```
Refinery::ExtensionHooks.register_stylesheets(  ['abc/foo'])
Refinery::ExtensionHooks.register_js_files(     ['abc/bar'])
```
## Customize the html
- App can override a `_cusomize_*` partial (e.g. `refinery/_customize_layout.html.erb` )
- Extension can register a partial (e.g. `Refinery::ExtensionHooks.register_html_partials(['refinery/customize_layout_foo'])`
- content_for blocks go in these files: (e.g. `content_for :head do` )
- yield & content_for? allow for customization (e.g. `yield :head` )
- an extension can add a scope 
  (e.g. `<%= render partial: 'refinery/extension_hooks', locals: {scope: :baz} %>` )
- Then other extensions can customize within that scope: (e.g. `Refinery::ExtensionHooks.register_html_partials(['refinery/customize_layout_stuff'], :baz)` )

### DIFF
--- a/core/app/assets/javascripts/glass/extension_assets.js.erb
+++ b/core/app/assets/javascripts/glass/extension_assets.js.erb
@@ -1,0 +1,5 @@
+<%
+  Refinery::ExtensionHooks.backend_js_files.each do |path|
+    require_asset(path)
+  end
+%>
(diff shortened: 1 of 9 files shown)


bricesanchez:
I'm closing this PR because it's not updated and i think we should rewrite the frontend of Refinery before thinking about adding new features for assets

PR #3213, opened 2016-07-10
Fix cms generator with heroku arg

bricesanchez:
This should fix `heroku` arg on cms generator

### DIFF
--- a/core/lib/generators/refinery/cms/cms_generator.rb
+++ b/core/lib/generators/refinery/cms/cms_generator.rb
@@ -81,8 +81,11 @@ ruby #{ENV['RUBY_VERSION'].inspect}
 
 # The Heroku gem allows you to interface with Heroku's API
 gem 'heroku'
+}
 
-group :production do
+      append_file 'Gemfile', "\ngroup :production do"
+
+      append_file 'Gemfile', %q{
   # Dragonfly's S3 Data Store extension allows you to use S3 assets (added for Heroku)
   gem 'dragonfly-s3_data_store'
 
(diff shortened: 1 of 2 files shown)


bricesanchez:
It looks like the error is still there or the generator build the new app with the rubygems source instead the github source: 

```
rails new my_new_application -m https://raw.githubusercontent.com/refinery/refinerycms/69572e9f31b967c78d2c444550984f7ccdc25c80/templates/refinery/installer.rb --heroku
```

parndt:
>[!] There was an error parsing `Gemfile`: syntax error, unexpected end-of-input, expecting keyword_end. Bundler cannot continue.

bricesanchez:
Yes, the last ´end´ does not print in the Gemfile, i don't understand why, could you help me to figure out?

parndt:
I am debugging it locally.

bricesanchez:
Thanks!

parndt:
I think it might have been a mistake adding `--heroku` into the test, because we don't want to be contacting third parties (especially when they'll set up a hosting account for us!) in automated testing.  See this build job for what I mean: https://travis-ci.org/refinery/refinerycms/jobs/145453641

bricesanchez:
Oh yes! We don't want this. And i think we don't want to automatically push to heroku even if it's not on test env.

parndt:
Well that's the way it's always worked for years now so I think that part is okay. It's just the testing of it that's not.

bricesanchez:
It's like an hotfix for me, if we remove --heroku, we don't know if the heroku script works. We don't want to push with no prompt message or conditional check. So it could be part of the 'heroku Q&A' i have to do!?

parndt:
Well then we need an additional argument like `--no-deploy` which overrides that part of the Heroku process?

PR #3216, opened 2016-07-13
enable reselection of resource after removal

NicholasKirchner:
I had a problem with the resource picker not allowing selection of a new resource after removing an already existing resource.


core/app/views/refinery/admin/_resource_picker.html.erb
```diff
@@ -13,7 +13,9 @@
 <%= f.hidden_field field %>
 
 <div>
-  <%= action_icon :add, insert_link, t('.name'), id: "current_resource_link_#{field}" unless resource %>
+  <div id="no_resource_selected_<%= field %>", style="<%= 'display: none' if resource %>">
```
bricesanchez: i will prefer the use of `content_tag` view helper, it's easier to read :

```ruby
<%= content_tag(:div, id: "no_resource_selected_#{field}", style: "display: none;" if resource) do %>
  <%= action_icon :add, insert_link, t('.name'), id: "current_resource_link_#{field}" %>
<% end %>
```

bricesanchez:
Thanks for your contribution, could you make this changes after @parndt review?

parndt:
Can you show us a before/after screenshot of what this is expected to change please?

NicholasKirchner:
The current flow is that we start from the following:
start_from_resource_present

From here, we click the remove button on the right side.  Without this change, there's nothing there:
without_pr_no_add_button

With this change, the add button re-appears:
with_pr_get_add_button

parndt:
Perfect, I can confirm the solution, thanks!

PR #3217, opened 2016-07-13
Refactor CustomScrubber to fix support of data-attr

bricesanchez:


pages/app/presenters/refinery/pages/section_presenter.rb
```diff
@@ -91,13 +91,16 @@ def initialize(tags, attributes)
...
+      end
 
-          super
-        end
+      def skip_node?(node)
+        node.text?
+      end
+
+      def scrub_attribute?(name)
+        !attributes.include?(name) && !(name =~ /\Adata-[\w-]+\z/).present?
```
parndt: We could use exclude? here

parndt: Could also use 
&& name !~ /\Adata-[\w-]+\z/

bricesanchez: true

kreker92:
It has success. Thanks!

Issue #3215, opened 2016-07-13
Sitemap multilanguage problem


sintro:
There are problem with sitemap generation when multiple locales are enabled.
The problem is in two places:
**1.**
https://github.com/refinery/refinerycms/blob/master/core/app/views/refinery/sitemap/index.xml.builder#L11
When this `page_url` (prepared as Hash) goes to `refinery.url_for(page_url)`, it has `locale` key with locale, set in my browser (which is always will be `:ru`), and when the `RouteFilter` tries to generate url for current ::I18n.locale, it uses **exactly** params[:locale] (look here https://github.com/refinery/refinerycms-i18n/blob/master/lib/refinery/i18n-filter.rb#L23). Probably, when sitemap.xml is called by search bots, there will be no :locale key in params, but I am not sure. The possible solution can be using modified line:
```
page.url.merge({:only_path => false, locale: locale})
```
This way, url_for will get modified params[:locale], as it is currently set buy `@locales.each do |locale|` loop. And the line`::I18n.locale = locale` can be removed.
**2.**
More serious problem takes place when the url_for gets **String** as `page_url`, which happens when the page with redirection is precessing (or even root page of site). In this case, the module `RoutingFilter::RefineryLocales` is not fired at all, so for all the locales in @locales array there will be absolutely identical urls. I don`t know how to solve this problem best, but there is at least stupid way to duplicate some code from here https://github.com/refinery/refinerycms-i18n/blob/master/lib/refinery/i18n-filter.rb#L30 to here https://github.com/refinery/refinerycms/blob/master/core/app/views/refinery/sitemap/index.xml.builder#L14, and this is obviously ugly solution. Any suggestions?

sintro:
I just solved the problem with less duplications, but I think it is still not the best solution. All changes I did in `sitemap\index.xml.builder`, here is how it looks now:
```
xml.instruct!

xml.urlset "xmlns" => "http://www.sitemaps.org/schemas/sitemap/0.9" do

  @locales.each do |locale|
    ::I18n.locale = locale
    ::Refinery::Page.live.in_menu.includes(:parts).each do |page|
     # exclude sites that are external to our own domain.
     page_url = if page.url.is_a?(Hash)
       # This is how most pages work without being overriden by link_url
       page.url.merge({:only_path => false, locale: locale})
     elsif page.url.to_s !~ /^http/
       # handle relative link_url 09addresses.
       raw_url = [request.protocol, request.host_with_port, page.url].join
[...]

PR #3218, opened 2016-07-14
Fixed sitemap generation for multiple frontend locales

sintro:
Fixes to sitemap/index.xml.builder for correct generation when multiple frontend locales are activated. 
More info in this https://github.com/refinery/refinerycms/issues/3215 issue.

### DIFF
--- a/core/app/views/refinery/sitemap/index.xml.builder
+++ b/core/app/views/refinery/sitemap/index.xml.builder
@@ -17,6 +17,8 @@ xml.urlset "xmlns" => "http://www.sitemaps.org/schemas/sitemap/0.9" do
          filter.around_generate({}) do
            raw_url
          end
+       else
+         raw_url
        end
      end
 
--- a/core/app/views/refinery/sitemap/index.xml.builder
+++ b/core/app/views/refinery/sitemap/index.xml.builder
@@ -8,10 +8,16 @@ xml.urlset "xmlns" => "http://www.sitemaps.org/schemas/sitemap/0.9" do
      # exclude sites that are external to our own domain.
      page_url = if page.url.is_a?(Hash)
        # This is how most pages work without being overriden by link_url
-       page.url.merge({:only_path => false})
+       page.url.merge({:only_path => false, locale: locale})
      elsif page.url.to_s !~ /^http/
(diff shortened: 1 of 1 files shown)


core/app/views/refinery/sitemap/index.xml.builder
```diff
@@ -8,10 +8,16 @@ xml.urlset "xmlns" => "http://www.sitemaps.org/schemas/sitemap/0.9" do
...
      elsif page.url.to_s !~ /^http/
        # handle relative link_url addresses.
-       [request.protocol, request.host_with_port, page.url].join
+       raw_url = [request.protocol, request.host_with_port, page.url].join
+       if (@locales.size > 1) && defined?(RoutingFilter::RefineryLocales)
+         filter = RoutingFilter::RefineryLocales.new
+         filter.around_generate({}) do
+           raw_url
+         end
+       end
```
parndt: It doesn't seem like `raw_url` is being returned anymore if this condition fails, as it's not being returned in the `else` case.

parndt:
Thanks for taking a look at this!

core/app/views/refinery/sitemap/index.xml.builder
sintro: You are right! Need to add else case to return raw_url without changes, or do some refactoring to get more effective correct code.

core/app/views/refinery/sitemap/index.xml.builder
```diff
@@ -17,6 +17,8 @@ xml.urlset "xmlns" => "http://www.sitemaps.org/schemas/sitemap/0.9" do
          filter.around_generate({}) do
            raw_url
          end
+       else
+         raw_url
```
sintro: Here is the fix, may be there are any better way to do it?

sintro: So, here is the fix, may be there are any better way to do it?

parndt:
Good enough for me - @bricesanchez ?

bricesanchez:
Could you squash and rebase your branch ?

parndt:
Done as a merge step

PR #3222, opened 2016-07-18
Prevent deployment to hosting when running tests

parndt:
This is for a change introduced with #3213


parndt:
@bricesanchez I want to merge this soon, and we can bring it back once we address a smart way of blocking the deployment aspect of the `--heroku` flag, what do you think about that?

parndt:
Of note is that the flag isn't doing anything that is being tested anyway because the tests pass on this PR

bricesanchez:
Waiting for CI then i will merge it.

PR #3230, opened 2016-07-21
Add Spanish locales

abelardogilm:
Updated the es.yml in line with the en.yml


bricesanchez:
Thanks @abelardogilm!

Issue #3087, opened 2016-01-08
Why is this admin nav repeating vertically?


jaredbeck:
![screen-shot-2016-01-06-at-6 23 56-pm](https://cloud.githubusercontent.com/assets/380809/12201245/1cd44d6a-b5f1-11e5-98e4-54f8fe31f007.png)

In the following chat on gitter, Brice asked me to create this issue. I don't have any more information.

> jaredbeck
> Jan 06 18:24
> Why is this admin nav repeating vertically?
> Using refinerycms-pages 3.0.0
> Is it drawn by refinerycms-core-3.0.0/app/views/refinery/admin/_menu.html.erb?
>
>
> gabyshev
> Jan 07 07:28
> @jaredbeck I faced the same issue when I was editing my locale files. Check them carefully and try to restart server.
>
>
> bricesanchez
> Jan 07 08:19
> @jaredbeck and @gabyshev it happens in dev env only, please fill an issue for reference

anitagraham:
I get this too, and I've assumed it's a result of rebuilding the side-nav after a dev  crash/restart.

parndt:
I think it's when you edit something in an extension and so the code reloads and something isn't clearing its cache properly!

michael-misshore:
I also would like to confirm the reported behavior. If I get some free time, I can look into submitting a pull request to fix this.

bricesanchez:
@michael-misshore :heart:

randallreedjr:
I am also seeing this, but it didn't start for me until I activated the `config.plugin_priority` setting in initializers/refinery/core.rb.

bricesanchez:
Any news @michael-misshore ?

michael-misshore:
@bricesanchez Unfortunately, I haven't been able to get to this particular issue. The last time I took a look at this, I had a tough time replicating it. Hopefully, I'll be able to take some time to investigate and ultimately get this corrected.

anitagraham:
@bricesanchez I get this all the time. It seems to be associated with restarting the server while continuing to use the same web page.

parndt:
I still believe that it's when you change some code in vendor/extensions/ or maybe a decorator and code isn't reloading.

Dreamjer:
Hi,
I see the plugins in users duplicated too :
![plugins_duplicate](https://cloud.githubusercontent.com/assets/11614574/14532463/7c6b7052-0261-11e6-8d75-e0d69ce6b479.jpg)
[...]

PR #3234, opened 2016-07-25
Fixed admin menu items duplicating after extensions code changing

sintro:
Fixing https://github.com/refinery/refinerycms/issues/3087#issuecomment-235118966 issue

### DIFF
--- a/core/lib/refinery/plugin.rb
+++ b/core/lib/refinery/plugin.rb
@@ -16,7 +16,7 @@ module Refinery
       plugin.class_name ||= plugin.name.camelize
 
       # add the new plugin to the collection of registered plugins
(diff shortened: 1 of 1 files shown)


parndt:
This is great, I really like this code :wink:

Once this passes CI I'll merge it.

sintro:
Nice, this method looks good to me also, because I think forcing the reloading or all the Refinery::Plugins after every Refinery::Plugin change is much less effective than just to filter an array one time.

parndt:
Absolutely agree 😄

parndt:
I've cherry picked this across to master too https://github.com/refinery/refinerycms/commit/23781a81db3e8ea4a16cfa06f36a15c4c4d2456c

PR #3094, opened 2016-02-02
POC : Improve nested pages admin tree

bricesanchez:
This PR tries to fix #2950.

It's not yet ready to be merged... 

...but let me know if it helps loading big page tree

@jaredbeck 
@parndt 

Without this fix it takes **25secs** to render this view with **127 pages** :
refinery-pages-4-master

With this fix it takes **10secs** to render this view with the same number of pages
refinery-pages-2-bugfix

That's better but not so good...

Note : I've translated pages using Globalize.

PR #3244, opened 2016-08-25
Added expended option to pages extension

sjoulbak:
Before this feature you could only enable expanding pages with children in the admin tree by setting the `config.auto_expand_admin_tree` to true. But you couldn't disable expanding for pages where expanding isn't necessary when `config.auto_expand_admin_tree` is set to true.

I added a boolean attribute `expand_in_admin_tree` (which default is true) to the pages entity so you can disable expanding for a page in the admin pages tree.

### DIFF
--- a/pages/app/controllers/refinery/admin/pages_controller.rb
+++ b/pages/app/controllers/refinery/admin/pages_controller.rb
@@ -100,7 +100,8 @@ module Refinery
         [
           :browser_title, :draft, :link_url, :menu_title, :meta_description,
           :parent_id, :skip_to_first_child, :show_in_menu, :title, :view_template,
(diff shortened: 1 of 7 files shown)


bricesanchez:
Thanks for you contribution!

It's a simple way to improve the `config.auto_expand_admin_tree` but the real problem is the time to load the page tree (#3094) and perhaps the fact that we aren't using cookies to keep pages expanded for a user. 

I don't think all admins would like to see the same list of expanded pages.

Any thoughts @refinery/collaborators ?

johanb:
@bricesanchez interesting, you might be right that expanding on a cookie per user would be better. @sjoulbak maybe we can look into #3094 together next week.

PR #3251, opened 2016-09-19
Enable responsive web design by adding  meta viewport in head tag

bricesanchez:
I didn't see it before because i always override the _head partial in my projects. But i think it should be here by default now.


parndt:
I'm 👎 on this because I don't think we should be deciding about whether users of Refinery have made a responsive design or not.  My belief is that anyone who makes their frontend responsive will add this anyway.

bricesanchez:
It makes sense and it's not so hard to add.

PR #3256, opened 2016-09-26
Improve layout HTML5 markup and menu cache

bricesanchez:
Improve HTML5 markup: According to http://html5doctor.com/the-main-element/
Add menu role
Add menu cache
Deprecate `cache_if`


parndt:
I'm :-1: on introducing more helpers

bricesanchez:
What could i do to make you agree with this PR?

parndt:
By introducing helpers it makes the code harder to find, and harder to override. I think helpers are an anti pattern 😬

parndt:
As this changes the HTML we need to release this as 3.1 or 4.0 because this may break web applications using the defaults with tageted CSS or JS

bricesanchez:
Could you review again this PR @parndt ?

bricesanchez:
friendly ping @parndt

core/app/views/refinery/_header.html.erb
```diff
@@ -1,4 +1,11 @@
-<h1 id='logo'>
-  <%= link_to Refinery::Core.site_name, refinery.root_path %>
-</h1>
-<%= Refinery::Pages::MenuPresenter.new(refinery_menu_pages, self).to_html %>
+<%= content_tag :div, id: "logo", itemscope: "",  itemtype: "http://schema.org/Organization" do %>
+  <%= link_to refinery.root_path, itemprop: "url" do %>
+    <%= content_tag :span, { itemprop: "name"} do %>
```
bricesanchez: Finally, after reading doc (http://schema.org/name) and examples (https://github.com/Shopify/Timber/blob/master/layout/theme.liquid#L169) i prefer to not add span surrounding the site name

bricesanchez:
:up: 😄

PR #3258, opened 2016-10-18
Remove deprecated heroku gem

bricesanchez:
```
Fetching: heroku-3.43.12.gem (100%)
 !    The `heroku` gem has been deprecated and replaced with the Heroku Toolbelt.
 !    Download and install from: https://toolbelt.heroku.com
 !    For API access, see: https://github.com/heroku/heroku.rb
```

### DIFF
--- a/core/lib/generators/refinery/cms/cms_generator.rb
+++ b/core/lib/generators/refinery/cms/cms_generator.rb
@@ -283,7 +283,7 @@ end
       generator_args << '--quiet' if self.options[:quiet]
       generator_args << '--skip-migrations' if self.options[:skip_migrations]
       Refinery::CoreGenerator.start generator_args
-      Refinery::AuthenticationGenerator.start generator_args if defined?(Refinery::AuthenticationGenerator)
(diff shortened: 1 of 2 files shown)


parndt:
We need this to be present, it's [used in the generator](https://github.com/refinery/refinerycms/blob/f4b3cf3eb674cbdd5097f302f028262be41f6a70/core/lib/generators/refinery/cms/cms_generator.rb#L161-L183), and I'm pretty sure it's had that deprecation message for a long time now.

bricesanchez:
It's provide by the toolbelt now and our guide explicit that we need to have the toolbelt.

parndt:
My major complaint is that if people specify  --heroku it won't automatically set things up, it'll fail.  Maybe we should abort if the Heroku command isn't available?

bricesanchez:
I've tried something, let me know if it's ok for you

bricesanchez:
friendly ping @parndt

bricesanchez:
🆙  😄

PR #3259, opened 2016-10-22
Add ability to display current used template in layout

bricesanchez:


bricesanchez:
When you have multiple templates and you use the same template on multiple pages, you will prefer to target a template name versus a page canonical slug for your CSS.

bricesanchez:
friendly ping @parndt

bricesanchez:
@parndt ?

Issue #3020, opened 2015-07-27
Pages preview doesn't post current body content unless preview button is clicked twice


steveforse:
Verified issue exists on latest commit to master branch (24b521700600d450b23d9dc6ce9c39514611f09d) by generating a new Refinery app.

Verified issue exists on Chrome and IE11.

Steps to reproduce:
1) Install latest Refinery
2) Create/edit page
3) Add new content to body
4) Click preview
5) Preview tab _DOES NOT_ contain latest content
6) Click preview again
7) Preview tab _DOES_ contain latest content

parndt:
A failing test case for this would be A+ excellent!

mlinfoot:
I thought I might help with this and learn some rspec along the way. I've managed to get most of it working, but I've hit an impasse and need some pointers. 

I have two tests, `it 'will not show the preview changes the first time' do` and `it 'will show the preview changes the second time' do`. The first one is working as expected, but the second one fails at the attempt to simulate the second Preview button click with:

    Failure/Error: window = window_opened_by do
    Capybara::WindowError:
      block passed to #window_opened_by opened 0 windows instead of 1

I've created a gist of the [code snippet here](https://gist.github.com/62bff3de3c8c651c7256). After the first `window_opened_by` is closed at line 33, I expected the browser focus to be back at the original new page window and the second `window_opened_by` at line 36 would click the Preview button again. Everything I've tried at this point has failed.

Thoughts? If you see a simple way to make this work, I'll be happy to make a PR.

anitagraham:
Marshall, do you really need the second test? We don't really want people to go to the second click on preview before it works.

What we need - I think - is the first test, but failing. I would suggest that your first test, reworded as below, is all that is needed.  And a fix.
I also split it from the original test which filled the title, and added a context to make that clear. I think.

````ruby
context 'when editing a body part` do
 it 'will show the changes in the preview' do
            visit refinery.admin_pages_path
            find('a', text: 'Add new page').trigger(:click)
        
            page.evaluate_script("WYMeditor.INSTANCES[1].html('<p>Some changes. What they will look like?</p>')")
            window = window_opened_by do
              click_button "Preview"
            end

            expect_window_without_content(preview_content, window: window)

            window.close
          end
  end
````

mlinfoot:
Much better, Anita. Thanks! If we want it to fail now, but pass when the issue is fixed, we should "expect_window_with_content". Like so:

    context 'when editing a body part` do
[...]

PR #3261, opened 2016-11-07
Fix preview button with WYMeditor (#3020)

aiomaster:


parndt:
@bricesanchez any chance you have some time to double check that this works for you too?

bricesanchez:
I confirm the fix. Thanks @aiomaster !

PR #3271, opened 2016-11-29
Fix #3218 regression on sitemap: don't duplicate locale in url

bricesanchez:
Fix #3218 regression: don't duplicate locale in url and refactor the builder to be more readable

Since this PR, i have problems on sitemap:
```xml
[...]
<url>
<loc>
http://localhost:3000/en/en/rate_requests/new
</loc>
<lastmod>2016-09-01</lastmod>
</url>
<url>
<loc>http://localhost:3000/en/forms</loc>
<lastmod>2016-09-01</lastmod>
</url>
<url>
<loc>http://localhost:3000/en/about-us</loc>
<lastmod>2016-09-01</lastmod>
</url>
<url>
<loc>http://localhost:3000/en/en/jobs</loc>
<lastmod>2016-09-01</lastmod>
</url>
<url>
<loc>http://localhost:3000/en/en</loc>
<lastmod>2016-07-20</lastmod>
</url>
[...]
```

After the fix: 

```xml
[...]
<url>
<loc>http://localhost:3000/en/rate_requests/new</loc>
<lastmod>2016-09-01</lastmod>
</url>
<url>
<loc>http://localhost:3000/en/forms</loc>
<lastmod>2016-09-01</lastmod>
</url>
<url>
<loc>http://localhost:3000/en/about-us</loc>
<lastmod>2016-09-01</lastmod>
</url>
<url>
<loc>http://localhost:3000/en/jobs</loc>
<lastmod>2016-09-01</lastmod>
</url>
<url>
<loc>http://localhost:3000/en</loc>
<lastmod>2016-07-20</lastmod>
</url>
[...]
```

### DIFF
--- a/core/app/views/refinery/sitemap/index.xml.builder
+++ b/core/app/views/refinery/sitemap/index.xml.builder
@@ -1,33 +1,39 @@
 xml.instruct!
 
 xml.urlset "xmlns" => "http://www.sitemaps.org/schemas/sitemap/0.9" do
-
   @locales.each do |locale|
     ::I18n.locale = locale
-    ::Refinery::Page.live.in_menu.includes(:parts).each do |page|
-     # exclude sites that are external to our own domain.
-     page_url = if page.url.is_a?(Hash)
-       # This is how most pages work without being overriden by link_url
-       page.url.merge({:only_path => false, locale: locale})
-     elsif page.url.to_s !~ /^http/
-       # handle relative link_url addresses.
-       raw_url = [request.protocol, request.host_with_port, page.url].join
-       if (@locales.size > 1) && defined?(RoutingFilter::RefineryLocales)
-         filter = RoutingFilter::RefineryLocales.new
-         filter.around_generate({}) do
-           raw_url
-         end
-       else
-         raw_url
-       end
-     end
 
-     # Add XML entry only if there is a valid page_url found above.
-     xml.url do
-       xml.loc refinery.url_for(page_url)
-       xml.lastmod page.updated_at.to_date
-     end if page_url.present? and page.show_in_menu?
+    ::Refinery::Page.live.in_menu.each do |page|
+      
+      # exclude sites that are external to our own domain.
+      if page.url.is_a?(Hash)
+        
+        # This is how most pages work without being overriden by link_url
+        page_url = page.url.merge({:only_path => false, locale: locale})
+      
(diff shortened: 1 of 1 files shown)


bricesanchez:
@sintro Could you test this PR ?

sintro:
Just checked, your patch works fine for my cases, so it breaks nothing. But I also did not have such problems, probably, you get it when the path of page has manually set /language/ part? I think now with this PR everything is fine!

One thing about this sitemap generation: what about filtering out the pages with that locales, which don't have translations for that language?

bricesanchez:
> Just checked, your patch works fine for my cases, so it breaks nothing. But I also did not have such problems, probably, you get it when the path of page has manually set /language/ part? I think now with this PR everything is fine!

The sitemap problem was on extensions urls, no pages url.

> One thing about this sitemap generation: what about filtering out the pages with that locales, which don't have translations for that language (for example, empty body or title or both)?

It's already done by the `in_menu` scope: 
https://github.com/refinery/refinerycms/blob/2c3b7c31314cc3fbcd425d7c461e14002bfda7fc/pages/app/models/refinery/page.rb#L121

sintro:
`It's already done by the in_menu scope`
Strangely, but I always get all of the pages with all of frontend_locales... Even for pages, which even does not have that locale symbol in Pages section in admin panel (so, there are no title and no body).

mhtremblay:
Is there something missing to merge this PR?

sintro:
@mhtremblay everything is fine. At least, new sitemap.xml.erb is better than older.

mhtremblay:
Can someone merge this PR plz?

core/app/views/refinery/sitemap/index.xml.builder
```diff
@@ -1,33 +1,39 @@
...
 
 xml.urlset "xmlns" => "http://www.sitemaps.org/schemas/sitemap/0.9" do
-
   @locales.each do |locale|
     ::I18n.locale = locale
-    ::Refinery::Page.live.in_menu.includes(:parts).each do |page|
-     # exclude sites that are external to our own domain.
-     page_url = if page.url.is_a?(Hash)
-       # This is how most pages work without being overriden by link_url
-       page.url.merge({:only_path => false, locale: locale})
```
mhtremblay: page.url.merge({**only_path: false**, locale: locale})

PR #3272, opened 2016-12-19
change warn to use rails logger

yule:
This allows backtrace_silencers to be able to ignore this error. It also behaves nicely on what is set as log level in the environment file.

Prior to this, having `config.log_level = :fatal` will still output this warning as it uses the kernel warn rather than the rails logging facility.

Given that refinery is designed to run in rails, this will be the case in all cases.


yule:
will add test....

parndt:
Thank you! Great work! 👍 

I've merged your work to 572fff862337edc0e091507c96bfa69027d3c597 and also made it possible to inject the logger, defaulting to the Rails logger in 46b5066d27ed60b4f9d0bb0abc89e09523100347

PR #3275, opened 2017-01-09
Update the updated_at field of the page itself when a page_part got updated

veger:
This fixes an issue with the sitemap: `updated_at` field does not get updated when the page (content) is updated.


parndt:
This is wonderful, thank you! 👍

PR #3278, opened 2017-01-29
Fix permitted params allowing new images to be uploaded

veger:
This PR fixes uploading a new image for existing images. I would have expected that with the original/wrong permitted params uploading the initial image would have failed (but it obviously does not).

I have been fiddling over 2 hours to get a test case for this PR, but my knowledge of rspec/features, capybara, and the combination of both is seriously lacking... So no test-case this time... :(

Manual test show that adding a new image still works, and uploading a new one does as well (I get an error when I upload an image with an different name, see #3279)


veger:
The failing does does not have anything to do with my PR. I had the same issue when trying to build a test-case. It is because the quiet_assets gem is trying to include Rails 5. It probably needs some version constraints (I 'solved it by removing the gem :P )

parndt:
Does it still work to upload _multiple_ images?

veger:
I did not know that it is possible to upload multiple images..! I'll check it out and report back

veger:
It still works when adding new images (they are all added as separate images), when modifying an image I do not seem to be able to select multiple (but I assume that is logical?)

parndt:
That's right, you can only upload multiple when you're adding _new_ images

bricesanchez:
Thanks @veger! It takes time but it's finally merged :D

PR #3284, opened 2017-03-03
Use selected_image parameter to select an image upon opening the dialog

veger:
Required for my 'image type' setting, but seems sensible to add to refinerycms as it is non-intrusive and also convenient for others.


parndt:
I've re-run the builds a couple of times and it just feels like the new tests are being a bit flaky.. probably they should be instructed to wait for something on the page before expecting their main assertion.

veger:
I fixed the flaky tests

When looking at build https://travis-ci.org/refinery/refinerycms/builds/333465780 it shows page content), it became apparent that sometimes the image id ranged from 4 to 6 instead of the expected 1 to 3 (no idea why).

Using the ids of the image objects fixes this, as the ids are now used dynamically instead of hard coded (expected) ids.

bricesanchez:
Thanks @veger! That's good for me. And you @parndt?

parndt:
Sounds like it just wasn't flushing the database and resetting the auto incrementing IDs properly

veger:
Yes indeed, and only sometimes... Resulting in the flaky tests.
Maybe some race condition in the (rspec) test cleanup?

bricesanchez:
Thank you @veger for your work !

PR #3287, opened 2017-03-27
Fixed variable name for custom editor options

bofrede:
The variable must be custom_visual_editor_boot_options.
See:
https://github.com/parndt/refinerycms-wymeditor/blob/master/app/assets/javascripts/refinery/boot_wym.js.erb#L5

And the documentation:
https://github.com/parndt/refinerycms-wymeditor#user-content-custom-selectable-styles-in-wymeditor


parndt:
Thank you @bofrede

PR #3296, opened 2017-07-04
Middleware loading fix for Rails 5

sjoulbak:
The routes should be reloaded after the initialise has finished.
Otherwise the initialise would be called before the middleware is loaded.

### DIFF
--- a/pages/lib/refinery/pages/engine.rb
+++ b/pages/lib/refinery/pages/engine.rb
@@ -44,6 +44,7 @@ module Refinery
 
       config.after_initialize do
         Refinery.register_extension(Refinery::Pages)
+        Rails.application.reload_routes!
       end
 
       protected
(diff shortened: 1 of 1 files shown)


bricesanchez:
It sounds obvious. 

Did you notice a better performance, load, bug fix after the change?

sjoulbak:
Firstly, I've updated the `nerds-and-company/feature/rails-5` branch so it can be merged when you guys agree with the changes.

The Devise engine sets the Warden configuration in the middleware stack. When the routes were reloaded the middleware stack was not completely finished thus the configuration was missing. To ensure all configuration is present this change moves the logic of reloading the routes to an after_initialize callback.

bricesanchez:
I'm ok with that, let's merge it in the main PR and wait for a global review. Thanks @sjoulbak!

PR #3298, opened 2017-07-06
Prevent homepage with locale in path to be duplicate

bricesanchez:


parndt:
Is it worth adding a spec to https://github.com/refinery/refinerycms/blob/master/core/spec/controllers/refinery/sitemap_controller_spec.rb ?

bricesanchez:
Yes, we should! I thinking of changing the way we generate sitemap in order to handle engines records

core/app/views/refinery/sitemap/index.xml.builder
```diff
@@ -17,7 +17,7 @@ xml.urlset "xmlns" => "http://www.sitemaps.org/schemas/sitemap/0.9" do
         # handle relative link_url addresses.
         raw_url = [request.protocol, request.host_with_port, page.url].join
         
-        if (@locales.size > 1) && !page.url.start_with?("/#{locale}/") && defined?(RoutingFilter::RefineryLocales)
+        if (@locales.size > 1) && !page.url.start_with?("/#{locale}") && !!defined?(RoutingFilter::RefineryLocales)
```
parndt: what does the `!!` add here?

PR #3299, opened 2017-07-08
Bugfix: canonical url now use current_frontend_locale instead of defa…

bricesanchez:
…ult_frontend_locale

According to google and Yoast: https://yoast.com/rel-canonical/


parndt:
So now this will always point to the current URL?

bricesanchez:
Yes, like in our blog posts engine. If we want to target other languages, we should use hreflang instead : https://yoast.com/hreflang-ultimate-guide/

PR #3300, opened 2017-07-08
Now we use redirect 301 to show pages

bricesanchez:
instead of 302. See
http://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to

### DIFF
--- a/pages/app/controllers/refinery/pages_controller.rb
+++ b/pages/app/controllers/refinery/pages_controller.rb
@@ -25,11 +25,11 @@ module Refinery
     #
     def show
       if should_skip_to_first_child?
-        redirect_to refinery.url_for(first_live_child.url) and return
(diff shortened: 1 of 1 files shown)


parndt:
What's the reason for adding permanent redirection for these cases?

bricesanchez:
>  Whether Google actually handles 302s properly is an open question. If a 302 is used instead of a 301, search engines might continue to index the old URL, and disregard the new one as a duplicate. Link popularity might be divided between the two urls, hurting search rankings. Search engines might figure out how to handle the 302, or they might not. Google spokespeople have said that they will treat a 302 as a 301 if they think the webmaster has made an error, but why take a chance, and what about other search engines?

https://www.hochmanconsultants.com/301-vs-302-redirect/

I think we should use 301 instead of 302 in order to be sure to be properly indexed on search engines

PR #3301, opened 2017-07-10
Textile doc to markdown

bricesanchez:
Guides are now written in Markdown for the new website
All guides has been correctly converted to markdown.
I've fixed some guides with updated content.
It will remain a complete QA to check if the guides are always relevant with new versions of Refinery. But we should do it in another PRs.

The current result can be seen here : http://www.refinerycms.com/guides


parndt:
@bricesanchez there's one file conflicting

bricesanchez:
It's rebased now!

PR #3122, opened 2016-02-27
Rails 5.1 support

bricesanchez:
Rails 5.1 support is under progress.

## TODO (updated 2017/08/11)
- [x] Add support for Rails 5.1
- [x] Fix SEO Meta, wait for PR merge: https://github.com/parndt/seo_meta/pull/31
- [x] Review of @parndt

PR #3302, opened 2017-07-11
Updating the SEO in the Pages Engine

sjoulbak:
Hello all,

I've been experiencing an issue with updating the `browser_title` and the `meta_description` of the `Refinery::Page` model in the `master` branch and in the #3122 PR (feature/rails-5). It's not possible to update only the `browser_title` or the `meta_description` after creating a new page. I think that this is caused by Globalize, but I cannot confirm this yet. 

It is possible to update the `browser_title` or the `meta_description` when there's one (or more) of the [other translated attributes](https://github.com/refinery/refinerycms/blob/master/pages/app/models/refinery/page.rb#L11) of the `Refinery::Page` model also edited.

I've changed the page meta data spec to confirm this issue, because we tried to [create a page at line 34](https://github.com/refinery/refinerycms/blob/master/pages/spec/models/refinery/page_meta_data_spec.rb#L34) and not update it.

Can anyone else also confirm this issue? And does someone have any thoughts why this could happen?

### DIFF
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -10,6 +10,8 @@ module Refinery
 
     translates :title, :menu_title, :custom_slug, :slug, :include => :seo_meta
 
+    after_save { translations.collect(&:save) }
+
     class Translation
       is_seo_meta
 
--- a/pages/spec/models/refinery/page_meta_data_spec.rb
+++ b/pages/spec/models/refinery/page_meta_data_spec.rb
@@ -4,7 +4,7 @@ require 'spec_helper'
 module Refinery
   describe Page do
     let(:page_title) { 'RSpec is great for testing too' }
-    let(:page) { subject.class.new(:title => page_title, :deletable => true)}
+    let(:page) { subject.class.create(title: page_title, deletable: true)}
 
     context 'responds to' do
       it 'meta_description' do


sjoulbak:
@bricesanchez I've created a [repo](https://github.com/sjoulbak/refinery-updating-page-seo-verification) to confirm this issue. But I've still got no clue why this is can't be done 🤔

sjoulbak:
Well.. It took some time, but I think it's a Globalize issue. 

We've a `Translation` class in the `Refinery::Page` model, which got `is_seo_meta` in it. The `Refinery::Page` translates some of its own attributes and also `:seo_meta`. The seo_meta attributes won't be updated when only the `browser_title` or the `meta_description` are given. This is because the translation won't be updated first.

So when we add one (or more) of the other translated attributes https://github.com/refinery/refinerycms/pull/3302#issue-242130763 and save the `Refinery::Page` model, than the seo_meta will be updated! This probably because the translation will be saved (which also got the seo_meta attributes) before the seo_meta . 

I've added an `after_save` in the `Refinery::Page` model which will save all translations after the page is saved. Now it's possible to update the seo_meta attributes without updating other attributes first. 

This is (for now) a temporary fix, because this could cause some performance issues when you've an application with allot of translations. 

I'm not sure, but https://github.com/globalize/globalize/pull/606 could be a reason why this is happening. 

Does anyone got any thoughts why it's not possible to update the seo_meta without the just added `after_save`?

bricesanchez:
I can confirm the bug with `globalize` on `5.1.0.beta1`. Your fix works but i don't know if it the best practices.

What do you think @parndt ?

parndt:
@sjoulbak if you do a git bisect (please) does the problem go away before that commit you referenced?

You can git bisect inside the globalize source code if you clone it locally, run the bisect inside the library, and target it from Refinery using a path source in the Gemfile.
If that doesn't make sense, let me know or ask @johanb and he should be able to help you with it!

bricesanchez:
Any news @sjoulbak?

bricesanchez:
I've dig into `seo_meta` gem and found that when we save a Page with `:browser_title` and `meta_description` fields to record, seo_meta raise a silent error: 

```
[41, 50] in /Users/bricesanchez/Sites/parndt/seo_meta/lib/seo_meta/instance_methods.rb
   41:       @seo_meta ||= ::SeoMetum.new :seo_meta_type => self.class.name
   42:     end
   43:
   44:     def save_meta_tags!
   45:       byebug
   46:       seo_meta.seo_meta_id ||= self.id unless seo_meta.persisted?
=> 47:       seo_meta.save
   48:     end
   49:   end
   50: end
(byebug) seo_meta.save
false
(byebug) seo_meta.errors
#<ActiveModel::Errors:0x007f9202a5faf0 @base=#<SeoMetum id: nil, seo_meta_id: 1, seo_meta_type: "Refinery::Page::Translation", browser_title: nil, meta_description: nil, created_at: nil, updated_at: nil>, @messages={:refinery_page_translation=>["must exist"]}, @details={:refinery_page_translation=>[{:error=>:blank}]}>
```

bricesanchez:
I've cherry-pick your commit into the Rais 5 branch #3122  (where it should be merged)
And i've fixed seo_meta: https://github.com/parndt/seo_meta/pull/31

PR #3308, opened 2017-08-15
Fixed page listing error: comparison of Fixnum with nil failed

bofrede:
When browsing at http://localhost:3000/refinery/pages
Pages translated to locales not mentioned in the Refinery::I18n.frontend_locales causes:
```
ArgumentError in Refinery::Admin::Pages#index
Showing .../refinerycms-pages-3.0.5/app/views/refinery/admin/pages/_page.html.erb where line #28 raised:
comparison of Fixnum with nil failed
```
This happens if you start translating into a language, but remove the locale from Refinery::I18n.frontend_locales if it is not ready at launch.


pages/app/views/refinery/admin/pages/_page.html.erb
```diff
@@ -25,7 +25,7 @@
 
     <% if Refinery::I18n.frontend_locales.many? %>
       <span class='locales'>
-        <% page.translations.sort_by{ |t| Refinery::I18n.frontend_locales.index(t.locale)}.each do |translation| %>
+        <% page.translations.sort_by{ |t| Refinery::I18n.locales.index(t.locale)}.each do |translation| %>
```
bricesanchez: I like this change but we should really change the locale switcher the way we add the locales edit links. Perhaps we should create a presenter. Thoughts @parndt ?

bofrede: It would also be nice if the webmaster could add and remove frontend languages, without requiring a deployment.

bricesanchez: I'm not sure if we want this, perhaps it's a good idea to ask a dev if we can add/remove a frontend language.

parndt:
Is it possible to have a test case for this patch please?

PR #3313, opened 2017-09-17
Add Rails version to generated migration

### LINKED ISSUE #3312: Extension generators produce migrations missing version
Rails (5?) requires migrations to specify the minimum version of Rails the migration was written for  otherwise it raises an error. Manually adding this version constraint before copying and running the migrations works fine.

```
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateVideosVideos < ActiveRecord::Migration[4.2]
/path/site/db/migrate/20170917165814_create_videos_videos.refinery_videos.rb:2:in `<top (required)>'
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateVideosVideos < ActiveRecord::Migration[4.2]
/path/site/db/migrate/20170917165814_create_videos_videos.refinery_videos.rb:2:in `<top (required)>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

sfcgeorge:
Rails no longer allows migrations without a Rails version specified. I chose 4.2 as that's the minimum rails version Refinery supports.

Fixes #3312

### DIFF
--- a/core/lib/generators/refinery/engine/templates/db/migrate/1_create_namespace_plural_name.rb.erb
+++ b/core/lib/generators/refinery/engine/templates/db/migrate/1_create_namespace_plural_name.rb.erb
@@ -1,4 +1,4 @@
-class Create<%= namespacing %><%= class_name.pluralize %> < ActiveRecord::Migration
+class Create<%= namespacing %><%= class_name.pluralize %> < ActiveRecord::Migration[4.2]
 
   def up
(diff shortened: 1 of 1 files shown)


bricesanchez:
Fixed with 899a58a3739dccbf19a5a2ea486024085c692793. Thanks @sfcgeorge for reporting this bug!

PR #3317, opened 2017-10-05
require refinery core nil_user class (needed for prod env)

bricesanchez:
I was trying to fix https://github.com/parndt/refinerycms-wymeditor/issues/47 

And in order to be able to start the dummy app on `RAILS_ENV=production`. i had to require refinery core nil_user class in the `authorisation_adapter`

Could you check this PR fix the problem?


parndt:
It's a good patch regardless

PR #3323, opened 2017-10-26
Now use https protocol for links to refinerycms.com

bricesanchez:


contributing.md
```diff
@@ -33,4 +33,4 @@ inclusion.
 
 ## Links ##
 
-See also the [Contributing to Refinery](http://refinerycms.com/guides/contributing-to-refinery) guide.
+See also the [Contributing to Refinery](/guides/contributing-to-refinery) guide.
```
parndt: this will 404 if you're viewing this on GitHub, which you probably are 😄

bricesanchez: Fixed

parndt: I don't see the change? The link should be: `https://www.refinerycms.com/guides/contributing-to-refinery`

bricesanchez: Sorry it's the only one i've missed.

Issue #3329, opened 2017-12-14
Preview Button in Firefox is not Working


enatividad:
Preview Button Does Not Work on Firefox.

I am using Rails 5.1.4 and Firefox Quantum Web Browser. I have already worked-around the issue in #3328 . However, Preview button is still not working (clicking it does not open a new tab).

bricesanchez:
Hi @enatividad, thanks for the report! I've found a quick fix for the FF problem, i don't really like it but it works : #3330

PR #3330, opened 2017-12-16
Bugfix/3328 3329/preview button

### LINKED ISSUE #3328: click on preview button in rails 5 disables all buttons
Hi,
I discovered a problem with rails 5 and the preview button.
The problem is, that rails 5 has the `disable_with`-feature enabled by default:
https://github.com/rails/rails/commit/3822a322a82a19a9341a21a0cb1e36653da09c46

This means, that a click on a submit button disables all submit buttons on a form. It is used to prevent double clicks, I think.
Problem is, that the preview button in refinery is a bit special and opens the preview in a new tab.
This is working ok, but if you switch back to the backend tab you can't click any of the form buttons anymore, cause they are disabled by that feature.
One way to workaround this is to disable this rails feature globally like:
`config.action_view.automatically_disable_submit_tag = false`
It is good enough for us, cause we have our own logic to prevent double clicks and that stuff.

Another way for refinery could be to add the `data: { disable_with: nil }` to every form button to prevent that behaviour. Problem is, that it could be usefull to have that feature available for the "save" and "save and continue" buttons.
Another option could be to trick around a little with javascript and try to remove the `disable` attribute from the input buttons, but it is hard (I tried a little), cause you have to do it after rails does the disable magic.

Any other suggestions?

bricesanchez:
This fix #3328 and #3329

### DIFF
--- a/core/app/views/refinery/admin/_continue_editing.html.erb
+++ b/core/app/views/refinery/admin/_continue_editing.html.erb
@@ -1,3 +1,4 @@
 <%= f.submit t('.save_and_continue_editing'),
-             :id => "submit_continue_button",
-             :class => "wymupdate button" if f.object.persisted? %>
+             id: 'submit_continue_button',
+             class: 'wymupdate button',
+             data: { disable_with: false } if f.object.persisted? %>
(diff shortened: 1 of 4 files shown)


bricesanchez:
You're welcome @parndt !

PR #3332, opened 2018-01-12
 rails asset:precompile attempts to connect to DB because

emaraschio:
Hello!

When I run `rake assets:precompile` in production mode it throws the error: `PG::ConnectionBad: could not connect to server: Connection refused`, so I added a check to see if the DB is connected.


emaraschio:
Great 😄

bricesanchez:
Thanks @emaraschio and @parndt !

PR #3336, opened 2018-01-16
Extract permitted params lists

matiasgarciaisaia:
So they can be easily extended by (...wait for it...) _extensions_.

### DIFF
--- a/pages/app/controllers/refinery/admin/page_parts_controller.rb
+++ b/pages/app/controllers/refinery/admin/page_parts_controller.rb
@@ -23,9 +23,13 @@ module Refinery
 
       protected
         def new_page_part_params
-          params.except(:part_index).permit(:title, :slug, :body, :locale)
+          params.except(:part_index).permit(permitted_new_page_part_params)
         end
 
+      private
+        def permitted_new_page_part_params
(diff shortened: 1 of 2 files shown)


pages/app/controllers/refinery/admin/page_parts_controller.rb
```diff
@@ -23,9 +23,12 @@ def destroy
 
       protected
         def new_page_part_params
-          params.except(:part_index).permit(:title, :slug, :body, :locale)
+          params.except(:part_index).permit(permitted_new_page_part_params)
         end
 
+        def permitted_new_page_part_params
```
bricesanchez: This method should be `private` like all other `permitted` methods in Refinery

matiasgarciaisaia:
👌 Fixed that! I can squash/rebase the commits before merging - made a different commit to help with reviewing.

bricesanchez:
@matiasgarciaisaia it will be perfect if you could do this, it's better for tracking changes in the git log! Thank you very much!

matiasgarciaisaia:
There! ☝️

bricesanchez:
Thank you @matiasgarciaisaia !

Issue #3340, opened 2018-01-29
heroku deploy not happening


eumario:
When running refinerycms myapp --heroku, all the gems get installed, but when it attempts to find the heroku toolchain, it doesn't think that it is installed, even though it shows this in the backtrace:

>== 20180129204362 RenameTablesToNewNamespace: migrated (0.0324s) ==============
>
>heroku-cli/6.15.18-fdf2097 (linux-x64) node-v9.4.0
>/var/lib/gems/2.3.0/gems/refinerycms-core-4.0.1/lib/generators/refinery/cms/cms_generator.rb:145:in `deploy_to_hosting?':   [ABORTING] Heroku Toolbelt is not installed. Please re-start the installer after installing at: (RuntimeError)
>https://devcenter.heroku.com/articles/heroku-command-line

As you can see, heroku --version is being executed, and is present, but the method that checks for it, is checking !!system("heroku --version"), when I run it in irb, it is showing that system() is returning true, so I don't understand why this isn't working.....

bricesanchez:
Hi @eumario, i can confirm the bug, it looks like the generator cannot see the heroku command even if the toolbelt is installed.

Any idea to fix this bug @parndt ?

parndt:
What is your Ruby version?

eumario:
eumario@eumario-S550CA:~$ ruby --version
ruby 2.3.1p112 (2016-04-26) [x86_64-linux-gnu]


Mario Steele
Ruby Developer
C# Developer
Java Developer

bricesanchez:
It happens for me too with ruby 2.4.2p198, heroku-cli/6.13.19, rbenv 1.0.0.

bricesanchez:
Any idea @parndt?

parndt:
Which version of the heroku CLI do you have installed?

eumario:
heroku-cli/6.15.22-3f1c4bd (linux-x64) node-v9.4.0


Mario Steele
Ruby Developer
C# Developer
Java Developer

bricesanchez:
@eumario This PR will fix your error: #3347

PR #3341, opened 2018-02-03
Only add puma when it's missing

parndt:
While looking at #3340 I got the following errors:

```
[!] There was an error parsing `Gemfile`: You cannot specify the same gem twice with different version requirements.
You specified: puma (~> 3.7) and puma (>= 0). Bundler cannot continue.

 #  from myapp/Gemfile:77
 #  -------------------------------------------
 #    gem 'rails_12factor'
 >    gem 'puma'
 #    gem 'pg'
 #  -------------------------------------------
```

Rails now uses puma by default. This should fix the error.


bricesanchez:
Thanks @parndt!

Issue #3342, opened 2018-02-05
rake db:migrate:reset failed because of "duplicate column name: custom_slug"


yadex205:
Hello.

When running `$ rake db:migrate:reset`, migration will be aborted because of `SQLite3::SQLException: duplicate column name: custom_slug: ALTER TABLE "refinery_pages" ADD "custom_slug" varchar`.

It can be reproduced just by creating project and moving into `$ refinerycms my_site && cd my_site`, and run db:migrate:reset `$ rake db:migrate:reset`. And also it occurs with MySQL.

I resolved the issue by rewriting `/db/migrate/xxxxxxxxxxxxxx_add_custom_slug_to_refinery_pages.rb` as following.

```diff
@@ -1,21 +1,16 @@
 # This migration comes from refinery_pages (originally 20140105190324)
 class AddCustomSlugToRefineryPages < ActiveRecord::Migration[4.2]
   def up
-    if page_column_names.exclude?('custom_slug')
-      add_column :refinery_pages, :custom_slug, :string
-    end
+    add_column :refinery_pages, :custom_slug, :string unless custom_slug_exists?
   end
 
   def down
-    if page_column_names.include?('custom_slug')
-      remove_column :refinery_pages, :custom_slug
-    end
+    remove_column :refinery_pages, :custom_slug if custom_slug_exists?
   end
 
   private
-  def page_column_names
-    return [] unless defined?(::Refinery::Page)
 
-    Refinery::Page.column_names.map(&:to_s)
+  def custom_slug_exists?
+    column_exists?(:refinery_pages, :custom_slug)
   end
 end
```

### My environment

* macOS 10.12
* Ruby 2.4.2p198
* RefineryCMS 4.0.1 + Rails 5.1.4
* MySQL 5.7.11, SQLite 3.16.0

yadex205:
> Hi @yadex205 ! Thanks for your contribution! Could you provide a pull request for your bugfix?

@bricesanchez I've created PR at https://github.com/refinery/refinerycms/pull/3343 !

parndt:
Fixed by #3343

PR #3343, opened 2018-02-07
Fix checking :custom_slug existence to use column_exists?

yadex205:
Fix the issue https://github.com/refinery/refinerycms/issues/3342


bricesanchez:
Thanks @yadex205 and @parndt!

PR #3347, opened 2018-02-27
Bugfix/3340/heroku deploy

### LINKED ISSUE #3340: heroku deploy not happening

bricesanchez:
This will fix #3340

### DIFF
--- a/core/lib/generators/refinery/cms/cms_generator.rb
+++ b/core/lib/generators/refinery/cms/cms_generator.rb
@@ -1,4 +1,5 @@
 require 'pathname'
+require 'mkmf'
 
 module Refinery
   class CmsGenerator < Rails::Generators::Base
@@ -242,7 +243,7 @@ end
     end
 
     def heroku_toolbelt_missing?
-      !!system("heroku --version")
+      find_executable("heroku").nil?
     end
 
     def manage_roadblocks!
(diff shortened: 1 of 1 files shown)


core/lib/generators/refinery/cms/cms_generator.rb
```diff
@@ -93,6 +94,10 @@ def append_heroku_gems!
 }.sub("{{production_gems}}", production_gems.join("\n  "))
     end
 
+    def current_ruby_version
+      (env_ruby_version = ENV['RUBY_VERSION']).present? ? env_ruby_version : RUBY_VERSION
```
parndt: for this you can use:
```ruby
def current_ruby_version
  ENV['RUBY_VERSION'].presence || RUBY_VERSION
end
```

bricesanchez: Your right i under use `.presence`, my bad! it's fixed now

PR #3350, opened 2018-03-02
Update `Refinery::Pages::Type.parts` to match the format introduced in 3535b906fa

joshmcarthur:
This pull request resolves an issue I noticed while registering custom types in `config/initializers/refinery/pages.rb`. 

In 3535b906fa, the format of specifying page parts for default parts changed from a string: `["Body"]`, to a hash with `title` and `slug` keys: `[{title: "Body", slug: "body"}]`. This change was made for the default page parts, but not for page parts specified against registered types.

I have included a simple regression test as a controller test. This issue can also be caused by adding a database migration to set a default view template (this is how I originally reproduced the issue).

As an interim fix, I have added the following code to the application I am working on:

``` ruby
module RefineryPagePartDiscovery
  def self.included(base)
    base.module_eval do
      class << self
        def default_parts_for(page)
          return default_parts unless page.view_template.present?
          types.find_by_name(page.view_template).parts
        end
      end
    end
  end
end
Refinery::Pages.send(:include, RefineryPagePartDiscovery)
```

This pull request makes this change (removing `titleize`), but also wraps the `parts` writer in some code that prints a deprecation warning for any applications passing any page parts in as array of strings - not that it would work anyway, but I wanted to be thorough.
Note that I've speculated that this kind of config change would be a minor version change.

I have also made adjustments to the default pages initializer that will be generated, and updated the markdown documentation in relation to `default_parts` to match the format used in the initializer.

### DIFF
--- a/doc/guides/2 - Refinery Basics/1 - Changing Page Parts.md
+++ b/doc/guides/2 - Refinery Basics/1 - Changing Page Parts.md
@@ -15,7 +15,7 @@ Within `config/initializers/refinery/pages.rb` you will see various default sett
 
 ```ruby
 # Configure global page default parts
-# config.default_parts = ["Body", "Side Body"]
+# config.default_parts = [{ title: "Body", slug: "body" }, { title: "Side Body", slug: "side_body" }]
 ```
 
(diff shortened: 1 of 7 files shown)


bricesanchez:
Thanks @joshmcarthur and @simi !

PR #3357, opened 2018-03-15
WIP: Feature/fragment cache

bricesanchez:


core/app/views/refinery/_header.html.erb
```diff
@@ -4,7 +4,7 @@
   <% end %>
 <% end %>
 
-<% cache [I18n.locale, 'main_menu', @page, Refinery::Page.fast_menu.map(&:updated_at).max] do %>
+<% cache [I18n.locale, current_refinery_user, @page, Refinery::Page.fast_menu] do %>
```
parndt: huh? There is no `current_refinery_user` for this frontend view

bricesanchez: Yes there is a `current_refinery_user`: https://github.com/refinery/refinerycms/blob/2c3b7c31314cc3fbcd425d7c461e14002bfda7fc/core/lib/refinery/application_controller.rb#L51

parndt: but it's going to be nil when you're logged out so your cache key will be `en__#{@page}_#{array of pages}` which seems like a strange key. What am I missing?

bricesanchez: Probably, I will change the way i generate the cache_key, i will also using the model `cache_key` method. But right now i've a `can't dump File` error in the specs and i don't know why.

pages/app/presenters/refinery/pages/content_page_presenter.rb
```diff
@@ -5,17 +5,22 @@ module Pages
...
         add_page_parts(page.parts) if page
         add_default_post_page_sections
       end
 
-    private
+      private
 
       def add_default_title_section(title)
-        add_section TitleSectionPresenter.new(:id => :body_content_title, :fallback_html => title)
+        Rails.cache.fetch([::I18n.locale, @current_refinery_user, @page, title]) do
```
parndt: previously these presenters had no association with `Rails`, nor to users or anything like that. I'm uncomfortable adding those when caching can happen in the layer that has these.

bricesanchez: I understand, how should we handle this?

parndt:
Is it something we need to worry about / maintain?

pages/app/presenters/refinery/pages/content_page_presenter.rb
parndt: Is it something we need to worry about / maintain?

bricesanchez: I think you're right, we should not worry about it right now:
* cache_pages_full_caching exists
* when a template is more complex, we don't fully use the `ContentPagePresenter`, so we can easily add Rais fragment cache per project.

After digging into this behavior, i think it's probably the admin pages#index who need a better cache: https://github.com/refinery/refinerycms/blob/master/pages/app/views/refinery/admin/pages/_page.html.erb

parndt:
Thanks Brice

PR #3358, opened 2018-03-20
Fix Reserved Word Documentation Typo

jcbantuelle:
The `:allow_slug` example was placed inside the quotes for the route's `to` action, instead of as a separate hash element. This minor text update fixes the example so that it matches what needs to be in the routes file.


bricesanchez:
Thanks @jcbantuelle !

jcbantuelle:
Thank you! We love the CMS, it's super easy to use and configure.

bricesanchez:
Thanks @jcbantuelle for your feedback! We try to do our best to improve the CMS. Don't hesitate to submit us another Pull Request, we need more contributors :D

PR #3359, opened 2018-03-21
Feature/admin page index cache

bricesanchez:
Displaying the page tree in the admin could be slow when there are a lot of pages in the Tree.

Using the fragment cache will help us to save time in rendering, it should fix this issue #2950 and this PR #3094

### DIFF
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -55,7 +55,7 @@ module Refinery
 
     # Docs for acts_as_nested_set https://github.com/collectiveidea/awesome_nested_set
     # rather than :delete_all we want :destroy
-    acts_as_nested_set :dependent => :destroy
+    acts_as_nested_set counter_cache: :children_count, dependent: :destroy
 
     friendly_id :custom_slug_or_title, FriendlyIdOptions.options
 
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -55,7 +55,7 @@ module Refinery
 
     # Docs for acts_as_nested_set https://github.com/collectiveidea/awesome_nested_set
     # rather than :delete_all we want :destroy
-    acts_as_nested_set counter_cache: :children_count, dependent: :destroy
+    acts_as_nested_set counter_cache: :children_count, dependent: :destroy, touch: true
 
     friendly_id :custom_slug_or_title, FriendlyIdOptions.options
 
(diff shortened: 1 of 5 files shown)


bricesanchez:
@parndt I think my specs fails because of this line: https://github.com/refinery/refinerycms/blob/c57b3838fc384d38ff12cb4c36664bd16dc06dc4/pages/app/models/refinery/page.rb#L384

When i modify it for: 

```ruby
if translations.empty? || translation_for(Refinery::I18n.default_frontend_locale, false).try(:slug).present?
```

It works. i don't know why i have this error only when the cache is true on the render partial.

bricesanchez:
friendly ping @parndt :)

pages/app/models/refinery/page.rb
```diff
@@ -73,6 +73,9 @@ def should_generate_new_friendly_id?
     before_destroy :deletable?
     after_save :reposition_parts!
 
+    after_save { update_all_descendants if descendants.any? }
+    after_move { update_all_descendants if descendants.any? }
```
parndt: I don't think you need the conditional, and possibly you can just use:

```ruby
after_save :update_all_descendants
after_move :update_all_descendants
```

pages/app/models/refinery/page.rb
```diff
@@ -378,11 +381,15 @@ def puts_destroy_help
...
-      if translations.empty? || translation_for(Refinery::I18n.default_frontend_locale, false).present?
+      if translations.empty? || translation_for(Refinery::I18n.default_frontend_locale, false).try(:slug).present?
         Refinery::I18n.default_frontend_locale
       else
         translations.first.locale
       end
     end
+
+    def update_all_descendants
+      self.descendants.update_all(updated_at: DateTime.now)
```
parndt: I think you can achieve this with the `touch` method http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-touch which will set the `updated_at` for you without any timezone worries

pages/app/models/refinery/page.rb
bricesanchez: You're right

pages/app/models/refinery/page.rb
bricesanchez: I completly forget this method, that's a lot cleaner!

bricesanchez:
@parndt i've updated my PR with your requested changes. Could you approve it if the build is green?

Issue #51, opened 2010-03-03
Ruby 1.9: indexer.rb:158: Invalid next


TylerRick:
I'm not sure how this is valid even in Ruby 1.8 (since I thought next could only be used in loops), but in 1.9 it won't even "compile" this:

<pre>
  def self.check_line_for_find_indexes(file_name, line)
    find_regexp = Regexp.new(/([A-Z]{1}[A-Za-z]+|self).(find){1}((_all){0,1}(_by_){0,1}([A-Za-z_]+))?\(([0-9A-Za-z"\':=>. \[\]{},]*)\)/)
    if matches = find_regexp.match(line)

      model_name, column_names, options = matches[1], matches[6], matches[7]

      if model_name == "self"
        model_name = File.basename(file_name).sub(/\.rb$/,'').camelize
        table_name = model_name.constantize.table_name
      else
        if model_name.respond_to?(:constantize)
          if model_name.constantize.respond_to?(:table_name)
            table_name = model_name.constantize.table_name
          else
            #puts "Unable to get the table_name for #{model_name.to_s}. it could be an ActiveResource"
            next
          end
        else
          #puts "Unable to constantize #{model_name.to_s}, if you are sure that #{model_name.to_s} is a valid class name, please file an issue on\nhttp://github.com/eladmeidar/rails_indexes\nPlease supply the relevant code as well, thank
          next
        end
      end
      #...
    end
  end
</pre>

<pre>
indexer.rb:18: Invalid next
indexer.rb:22: Invalid next
indexer.rb: compile error (SyntaxError)
</pre>

PR #3360, opened 2018-03-24
Switch from Globalize to Mobility

bricesanchez:
In order to translate our content we have been using a library called
Globalize.  This has worked really well for us up until now, however a
new library has been released called Mobility which seems like it would
work better for our case.

This commit switches us from Globalize to Mobility in conjunction with
the release of refinerycms-i18n version 5.0.0.

This PR is stable, it needs code review but we have a green build: 

Big thanks to @shioyama !

I will merge this PR after : 
* [x] code review
* [x] the release of `mobility` 0.6.0: https://github.com/shioyama/mobility/releases
* [x] decide if `mobility` should be a dependency of `refinerycms-i18n`
* [x] release a new version of `refinerycms-i18n`, we will probably bump to 5.0

### DIFF
--- a/core/refinerycms-core.gemspec
+++ b/core/refinerycms-core.gemspec
@@ -23,7 +23,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version = Refinery::Version.required_ruby_version
 
   s.add_dependency 'refinerycms-i18n',            ['~> 4.0', '>= 4.0.0']
-  s.add_dependency 'awesome_nested_set',          ['~> 3.0', '>= 3.0.0']
   s.add_dependency 'railties',                    rails_version
   s.add_dependency 'activerecord',                rails_version
   s.add_dependency 'actionpack',                  rails_version
--- a/core/refinerycms-core.gemspec
+++ b/core/refinerycms-core.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = Refinery::Version.required_ruby_version
 
-  s.add_dependency 'refinerycms-i18n',            ['~> 4.0', '>= 4.0.0']
+  s.add_dependency 'refinerycms-i18n',            ['~> 5.0', '>= 5.0.0']
   s.add_dependency 'railties',                    rails_version
   s.add_dependency 'activerecord',                rails_version
   s.add_dependency 'actionpack',                  rails_version
--- a/images/db/migrate/20150430171341_translate_refinery_images.rb
+++ b/images/db/migrate/20150430171341_translate_refinery_images.rb
@@ -1,22 +1,20 @@
 class TranslateRefineryImages < ActiveRecord::Migration[4.2]
-  def self.up
-    begin
-      ::Refinery::Image.create_translation_table!({
-        image_alt: :string,
-        image_title: :string
-      }, {
-        :migrate_data => true
-      })
-    rescue NameError
-      warn "Refinery::Image was not defined!"
-    end
-  end
+  def change
+    create_table :refinery_image_translations do |t|
 
-  def self.down
-    begin
-      Refinery::Image.drop_translation_table! migrate_data: true
-    rescue NameError
-      warn "Refinery::Image was not defined!"
+      # Translated attribute(s)
+      t.string :image_alt
+      t.string :image_title
+
+      t.string  :locale, null: false
+      t.integer :refinery_image_id, null: false
+
+      t.timestamps null: false
     end
+
+    add_index :refinery_image_translations, :refinery_image_id, name: :index_refinery_image_translations_on_refinery_image_id
+    add_index :refinery_image_translations, :locale, name: :index_refinery_image_translations_on_locale
+    add_index :refinery_image_translations, [:refinery_image_id, :locale], name: :index_2f245f0c60154d35c851e1df2ffc4c86571726f0, unique: true
+
   end
 end
--- a/images/db/migrate/20150430171341_translate_refinery_images.rb
+++ b/images/db/migrate/20150430171341_translate_refinery_images.rb
@@ -12,9 +12,7 @@ class TranslateRefineryImages < ActiveRecord::Migration[4.2]
       t.timestamps null: false
     end
 
-    add_index :refinery_image_translations, :refinery_image_id, name: :index_refinery_image_translations_on_refinery_image_id
     add_index :refinery_image_translations, :locale, name: :index_refinery_image_translations_on_locale
     add_index :refinery_image_translations, [:refinery_image_id, :locale], name: :index_2f245f0c60154d35c851e1df2ffc4c86571726f0, unique: true
-
   end
 end
--- a/images/refinerycms-images.gemspec
+++ b/images/refinerycms-images.gemspec
@@ -20,7 +20,6 @@ Gem::Specification.new do |s|
   s.test_files        = `git ls-files -- spec/*`.split("\n")
 
   s.add_dependency 'refinerycms-dragonfly',   '~> 1.0'
-  s.add_dependency 'mobility',               '~> 0.5'
   s.add_dependency 'activemodel-serializers-xml', '~> 1.0', '>= 1.0.1'
   s.add_dependency 'refinerycms-core',        version
 
--- a/images/refinerycms-images.gemspec
+++ b/images/refinerycms-images.gemspec
@@ -19,9 +19,8 @@ Gem::Specification.new do |s|
   s.files             = `git ls-files`.split("\n")
   s.test_files        = `git ls-files -- spec/*`.split("\n")
 
-  s.add_dependency 'refinerycms-dragonfly',   '~> 1.0'
-  s.add_dependency 'activemodel-serializers-xml', '~> 1.0', '>= 1.0.1'
-  s.add_dependency 'refinerycms-core',        version
+  s.add_dependency 'refinerycms-core', version
+  s.add_dependency 'refinerycms-dragonfly', '~> 1.0'
 
   s.required_ruby_version = Refinery::Version.required_ruby_version
 
--- a/images/refinerycms-images.gemspec
+++ b/images/refinerycms-images.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
   s.test_files        = `git ls-files -- spec/*`.split("\n")
 
   s.add_dependency 'refinerycms-dragonfly',   '~> 1.0'
-  s.add_dependency 'globalize',               ['>= 5.1.0.beta1', '< 5.2']
+  s.add_dependency 'mobility',               '~> 0.5'
   s.add_dependency 'activemodel-serializers-xml', '~> 1.0', '>= 1.0.1'
   s.add_dependency 'refinerycms-core',        version
 
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -10,7 +10,7 @@ module Refinery
     extend Mobility
     translates :title, :menu_title, :custom_slug, :slug, :browser_title, :meta_description
 
-    after_save { translations.in(Mobility.locale).seo_meta.save! }
+    after_save { translations.in_locale(Mobility.locale).seo_meta.save! }
 
     class Translation
       is_seo_meta
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -11,6 +11,8 @@ module Refinery
     extend Mobility
     translates :title, :menu_title, :custom_slug, :slug, :browser_title, :meta_description
 
+    default_scope { i18n }
+
     class Translation
       is_seo_meta
     end
@@ -19,14 +21,14 @@ module Refinery
       def self.options
         # Docs for friendly_id https://github.com/norman/friendly_id
         friendly_id_options = {
-          use: [:reserved],
+          use: [:mobility, :reserved],
           reserved_words: Refinery::Pages.friendly_id_reserved_words
         }
         if ::Refinery::Pages.scope_slug_by_parent
           friendly_id_options[:use] << :scoped
           friendly_id_options.merge!(scope: :parent)
         end
-        friendly_id_options[:use] << :mobility
+        # friendly_id_options[:use] << :mobility
         friendly_id_options
       end
     end
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -47,7 +47,7 @@ module Refinery
 
     has_many :parts, -> {
       scope = order('position ASC')
-      scope = scope.includes(:translations) if ::Refinery::PagePart.respond_to?(:translation_class)
+      scope = scope.includes(:translations) if ::Refinery::PagePart.respond_to?(:mobility)
       scope
     },       :foreign_key => :refinery_page_id,
              :class_name => '::Refinery::PagePart',
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -10,6 +10,8 @@ module Refinery
     extend Mobility
     translates :title, :menu_title, :custom_slug, :slug, :browser_title, :meta_description
 
+    after_save { translations.in(Mobility.locale).seo_meta.save! }
+
     class Translation
       is_seo_meta
     end
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -11,8 +11,6 @@ module Refinery
     extend Mobility
     translates :title, :menu_title, :custom_slug, :slug, :browser_title, :meta_description
 
-    default_scope { i18n }
-
     class Translation
       is_seo_meta
     end
@@ -48,8 +46,8 @@ module Refinery
     friendly_id :custom_slug_or_title, FriendlyIdOptions.options
 
     has_many :parts, -> {
-      scope = order('position ASC')
-      scope = scope.includes(:translations) if ::Refinery::PagePart.respond_to?(:mobility)
+      scope = ::Refinery::PagePart.respond_to?(:mobility) ? i18n.includes(:translations) : all
+      scope = scope.order('position ASC')
       scope
     },       :foreign_key => :refinery_page_id,
              :class_name => '::Refinery::PagePart',
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -40,7 +40,7 @@ module Refinery
           friendly_id_options[:use] << :scoped
           friendly_id_options.merge!(scope: :parent)
         end
-        # friendly_id_options[:use] << :mobility
+
         friendly_id_options
       end
     end
--- a/pages/app/models/refinery/page.rb
+++ b/pages/app/models/refinery/page.rb
@@ -9,7 +9,7 @@ module Refinery
     extend FriendlyId
 
     extend Mobility
-    translates :title, :menu_title, :custom_slug, :slug, :include => :seo_meta
+    translates :title, :menu_title, :custom_slug, :slug, :browser_title, :meta_description
 
     attribute :title
     attribute :menu_title
@@ -20,10 +20,6 @@ module Refinery
     
     class Translation
       is_seo_meta
-
-      def self.seo_fields
-        ::SeoMeta.attributes.keys.map{ |a| [a, :"#{a}="]}.flatten
-      end
     end
 
     class FriendlyIdOptions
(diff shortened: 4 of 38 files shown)


pages/app/models/refinery/page.rb
```diff
@@ -36,7 +26,7 @@ def self.options
           friendly_id_options[:use] << :scoped
           friendly_id_options.merge!(scope: :parent)
         end
-        friendly_id_options[:use] << :globalize
+        friendly_id_options[:use] << :mobility
```
shioyama: Can you try moving this line up so it comes *before* `:scoped` in the array to `friendly_id_options[:use]` ?

bricesanchez: I've tried but it sadly does not change anything 😢

pages/app/models/refinery/page.rb
```diff
@@ -61,7 +49,7 @@ def should_generate_new_friendly_id?
 
     has_many :parts, -> {
       scope = order('position ASC')
-      scope = scope.includes(:translations) if ::Refinery::PagePart.respond_to?(:translation_class)
+      scope = scope.includes(:translations) if ::Refinery::PagePart.respond_to?(:mobility)
```
shioyama: I think rather than use a `default_scope` on everything, you could add it here:

```ruby
scope = ::Refinery::PagePart.respond_to?(:mobility) ? i18n.includes(:translations) : all
scope = scope.order('position ASC')
scope
```

pages/app/models/refinery/page.rb
```diff
@@ -8,35 +8,27 @@ module Refinery
...
+    extend Mobility
+    translates :title, :menu_title, :custom_slug, :slug, :browser_title, :meta_description
 
-    attribute :title
-    attribute :menu_title
-    attribute :custom_slug
-    attribute :slug
-
-    after_save { translations.collect(&:save) }
+    default_scope { i18n }
```
shioyama: This might fix some issues, but ideally it would be better not to use it if possible.

pages/lib/refinery/pages/finder.rb
```diff
@@ -26,15 +26,15 @@ def initialize(conditions)
...
-      def with_globalize
-        globalized_conditions = {:locale => ::Globalize.locale.to_s}.merge(conditions)
-        translations_conditions = translations_conditions(globalized_conditions)
+      def with_mobility
+        mobility_conditions = {:locale => ::Mobility.locale.to_s}.merge(conditions)
+        translations_conditions = translations_conditions(mobility_conditions)
 
         # A join implies readonly which we don't really want.
-        Page.where(globalized_conditions).
+        Page.where(mobility_conditions).
```
shioyama:
```ruby
Page.i18n.where(mobility_conditions)
```

pages/app/models/refinery/page.rb
```diff
@@ -376,9 +364,9 @@ def puts_destroy_help
     end
 
     def slug_locale
-      return Globalize.locale if translation_for(Globalize.locale, false).try(:slug).present?
+      return Mobility.locale if slug_backend.read(Mobility.locale).present?
```
shioyama: I'm not quite sure about the context, but `slug_backend.read(Mobility.locale)` is basically the same as `slug` here. This will build the translation for that locale if none exists yet.

With Globalize you have `translation_for(Globalize.locale, false)` which will avoid building the translation for the current locale if none yet exists. This is a bit different.

Mobility does expose a method `translation_for` [on the backend](https://github.com/shioyama/mobility/blob/b42686f65fa3eb98ed1c9bded6ded8a1c816f2d0/lib/mobility/backends/active_record/table.rb#L162-L166), but it has no option to override building an empty translation. You shouldn't really need that option though because [Mobility anyway removes any empty translations before saving](https://github.com/shioyama/mobility/blob/b42686f65fa3eb98ed1c9bded6ded8a1c816f2d0/lib/mobility/backends/active_record/table.rb#L168-L177), so I think here you should be able to safely replace `slug_backend.read(Mobility.locale)` with just `slug`, and `slug_backend.read(Refinery::I18n.default_frontend_locale)` with just `slug(locale: Refinery::I18n.default_frontend_locale)`.

You don't really need `present?` either since the [Presence plugin](https://github.com/shioyama/mobility/blob/b42686f65fa3eb98ed1c9bded6ded8a1c816f2d0/lib/mobility/plugins/presence.rb) converts any blank strings to `nil` for you already.

shioyama:
I haven't tried running this PR locally, but just looking at the code I noticed a few places where perhaps the `i18n` scope was missing. Maybe that will get some more tests passing... not really sure.

bricesanchez:
@shioyama Thank you for your review! I will take a look on this today :)

pages/lib/refinery/pages/finder.rb
```diff
@@ -51,7 +51,7 @@ def translations_conditions(original_conditions)
         translations_conditions = {}
         original_conditions.keys.each do |key|
           if translated_attributes.include? key.to_s
-            translations_conditions["#{Page.translation_class.table_name}.#{key}"] = original_conditions.delete(key)
+            translations_conditions["#{Page::Translation.table_name}.#{key}"] = original_conditions.delete(key)
```
shioyama: Do you even need to do this? Both Globalize and Mobility (with i18n scope) should seemlessly handle translated attributes as if they were normal attributes. What Refinery is doing here is basically what Mobility is doing [here](https://github.com/shioyama/mobility/blob/b42686f65fa3eb98ed1c9bded6ded8a1c816f2d0/lib/mobility/backends/active_record/table/query_methods.rb#L97), which it automatically does when you call `Post.i18n.where(title: "foo")` etc.

shioyama: ;tldr; I don't think you should really need the `translations_conditions` with either Globalize or Mobility.

images/db/migrate/20150430171341_translate_refinery_images.rb
```diff
@@ -1,22 +1,20 @@
...
-      Refinery::Image.drop_translation_table! migrate_data: true
-    rescue NameError
-      warn "Refinery::Image was not defined!"
+      t.string  :locale, null: false
+      t.integer :refinery_image_id, null: false
+
+      t.timestamps null: false
     end
+
+    add_index :refinery_image_translations, :refinery_image_id, name: :index_refinery_image_translations_on_refinery_image_id
```
shioyama: This index has been removed in the generators in the latest version of Mobility, see shioyama/mobility#198. I think you don't need this one (since it's covered by the last index below on the combo of `refinery_image_id` and `locale`).

pages/lib/refinery/pages/finder.rb
bricesanchez: You probably right, it's the first time i dig into the i18n logic of Refinery, there is probably legacy things we don't ever need now.

bricesanchez:
Thanks @shioyama for your help! Now there is just 11 fail specs.

shioyama:
Removing `translations_conditions` didn't work? That's strange...

bricesanchez:
> Removing translations_conditions didn't work? That's strange...

I've revert for now it because it causes more fail specs.
Most of my problems was caused by `should_generate_new_friendly_id?`

I have to clean up my PR too.

bricesanchez:
Hi @parndt and @shioyama ! The last 2 failing specs of this PR are caused by seo_meta and mobility, do you have any idea to fix them?

pages/spec/features/refinery/pages_spec.rb
```diff
@@ -425,12 +425,12 @@ def standard_page_menu_items_exist?
         let(:nested_page_slug) { 'nested_page' }
 
         let!(:nested_page) do
-          Globalize.fallbacks = [:ru]
-          _page = Globalize.with_locale(:en) {
+          Mobility.new_fallbacks[:ru]
```
shioyama: Do you need this? I'm planning to deprecate `new_fallbacks`. I don't think this does anything here.

pages/app/models/refinery/page.rb
```diff
@@ -164,20 +150,20 @@ def nullify_duplicate_slugs_under_the_same_parent!
     end
 
     def translated_to_default_locale?
-      persisted? && translations.any?{ |t| t.locale == Refinery::I18n.default_frontend_locale}
+      persisted? && translations.any?{ |t| t.locale.to_sym == Refinery::I18n.default_frontend_locale}
```
shioyama: Nice catch! Mobility doesn't convert translation locales to symbols like Globalize does.

shioyama: Added [a note about this](https://github.com/shioyama/mobility/wiki/Migrating-from-Globalize#translation-locale-is-a-string-not-a-symbol) to the Mobility wiki page on migrating from Globalize.

pages/spec/features/refinery/pages_spec.rb
shioyama: If you want to just set the global fallbacks for this one spec, couldn't you set `I18n.fallbacks` instead?

images/refinerycms-images.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
   s.test_files        = `git ls-files -- spec/*`.split("\n")
 
   s.add_dependency 'refinerycms-dragonfly',   '~> 1.0'
-  s.add_dependency 'globalize',               ['>= 5.1.0.beta1', '< 5.2']
+  s.add_dependency 'mobility',               '~> 0.5'
```
shioyama: Isn't it a bit strange to have the mobility dependency in `refinerycms-images` ? I had trouble tracking down where it was. I would expect it to be in `refinerycms-i18n`.

shioyama:
@bricesanchez The problem with seo_meta has to do with how Mobility handles dirty attributes, which is slightly different from Globalize.

Refinery has seo_meta on the model's translations, but when the seo_meta changes, Mobility picks it up as a change on the model, not the translation. So it doesn't save the translation, and loses the association to the seo meta, which is why it comes out `nil`.

I'm trying to figure out if/how Mobility should handle this.

shioyama:
So after a lot of digging through Mobility, Globalize, SeoMeta, and Rails, I understand the issue here and have a quick fix, although not terribly elegant.

Just add an `after_save` callback to the `Refinery::Pages::Core::BaseModel` class like this:

```ruby
after_save { translations.find { |t| t.locale == Mobility.locale.to_s }.seo_meta.save! }
```

This will force the model to save seo meta for the translation in the current locale. I find this fixes specs.

In shioyama/mobility#202, I'm adding an extension to the `translations` association which will make this a bit cleaner, so you will be able to do:

```ruby
after_save { translations.in(locale).seo_meta.save! }
```

if you don't mind waiting until I release 0.6.0 (should be sometime this week), this would be nicer.

The problem, in a nutshell, is that Mobility detects that there is a change in the translation and marks it as a change to a virtual attribute `meta_description_en`. For a normal column on the translations table, this works fine, but in this case SeoMeta is delegating `meta_description` setter/getter to its association, so this fails to trigger an update to the translation, and thus the SeoMeta also does not get updated.

By using `after_save` on the parent model, we just skip the translation entirely and force the seo meta to get updated.

I don't think I want to do anything special in Mobility to handle this case since it's actually quite tricky; I don't think most people will have delegated methods on their model translation class.

images/refinerycms-images.gemspec
bricesanchez: You're right, it's a bit strange, i think it's because `refinerycms-i18n` was not a dependency of `refinerycms-core` a long time ago.

We could add a dependency of `mobility` gem in `refinerycms-i18n`.

What do you think @parndt?

shioyama:
I've merged shioyama/mobility#202, so you can point to master for now. I'll release a new version soon.

bricesanchez:
Thank you one more time @shioyama !

Your help is awesome! We finally have a green build!

I've pinged @parndt to ask for a code review and take some decisions: https://github.com/refinery/refinerycms/pull/3360#issue-177220809

I'm really happy to use Mobility!

shioyama:
Great to hear we're at green! Just one thing: I'm changing the `in` method to `in_locale`, based on discussion in shioyama/mobility#202. I'll merge the change to master soon, and you'll have to update your code here.

shioyama:
Ok, I've merged the rename to `in_locale`, if you point to the latest master that method name should work.

parndt:
@shioyama thank you for all the assistance, Chris!

images/refinerycms-images.gemspec
parndt: Unless I'm mistaken it's because `refinerycms-images` doesn't depend on `refinerycms-i18n` so in order for images to be translated it has to depend on the translation library?

pages/lib/refinery/pages/finder.rb
parndt: @bricesanchez do you want to try removing it? I believe it was just a performance optimisation to get everything loaded in the single query.

images/refinerycms-images.gemspec
bricesanchez: > Unless I'm mistaken it's because refinerycms-images doesn't depend on refinerycms-i18n

You're right

> so in order for images to be translated it has to depend on the translation library?

You're right but `refinerycms-core` depends to `refinerycms-i18n`:
https://github.com/refinery/refinerycms/blob/90a44fc0ed08f541d285e0973670f4c21850c593/core/refinerycms-core.gemspec#L25

And `refinerycms-images` depends to `refinerycms-core`.

So what should we do @parndt?

parndt: I'd go with @shioyama's suggestion

bricesanchez: > ah well in that case this library shouldn't duplicate dependencies

It does not duplicate dependencies because `refinerycms-i18n` has no dependencies to `mobility` nor `globalize`: https://github.com/refinery/refinerycms-i18n/blob/master/refinerycms-i18n.gemspec#L19-L20

 even if we call Mobility in this gem: https://github.com/refinery/refinerycms-i18n/pull/76/files

So i will create a dependency to `mobility` in `refinerycms-i18n` like it should be.

pages/lib/refinery/pages/finder.rb
bricesanchez: @parndt i had a lot of failling specs (108 fails) when i tried to removed this method: 

```
  108) Crudify xhr_paging performs ajax paging of index
       Failure/Error: stmt = SQLite3::Statement.new( self, sql )

       ActiveRecord::StatementInvalid:
         SQLite3::SQLException: no such column: refinery_pages.locale: SELECT  "refinery_pages".* FROM "refinery_pages" INNER JOIN "refinery_page_translations" "translations_refinery_pages" ON "translations_refinery_pages"."refinery_page_id" = "refinery_pages"."id" INNER JOIN "refinery_page_translations" ON "refinery_page_translations"."refinery_page_id" = "refinery_pages"."id" AND "refinery_page_translations"."locale" = 'en' WHERE "refinery_pages"."locale" = 'en' AND "refinery_pages"."parent_id" IS NULL AND "refinery_page_translations"."slug" = ? ORDER BY "refinery_pages"."id" ASC LIMIT ?
       # /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/sqlite3-1.3.13/lib/sqlite3/database.rb:91:in `initialize'
       # /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/sqlite3-1.3.13/lib/sqlite3/database.rb:91:in `new'
       # /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/sqlite3-1.3.13/lib/sqlite3/database.rb:91:in `prepare'
       # /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/2.5.0/monitor.rb:226:in `mon_synchronize'
       # /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/2.5.0/monitor.rb:226:in `mon_synchronize'
       # ./pages/lib/refinery/pages/finder.rb:122:in `parent_page'
       # ./pages/lib/refinery/pages/finder.rb:106:in `find'
       # ./pages/lib/refinery/pages/finder.rb:83:in `find'
       # ./pages/lib/refinery/pages/finder.rb:149:in `find'
       # ./pages/lib/refinery/pages/finder.rb:9:in `by_path_or_id'
       # ./pages/app/models/refinery/page.rb:80:in `find_by_path_or_id'
       # ./pages/app/controllers/refinery/pages_controller.rb:110:in `call'
       # ./pages/app/controllers/refinery/pages_controller.rb:75:in `find_page'
       # /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/request_store-1.4.1/lib/request_store/middleware.rb:19:in `call'
       # /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/2.5.0/webrick/httpserver.rb:140:in `service'
       # /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/2.5.0/webrick/httpserver.rb:96:in `run'
       # /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/2.5.0/webrick/server.rb:307:in `block in start_thread'
       # ------------------
       # --- Caused by: ---
       # SQLite3::SQLException:
       #   no such column: refinery_pages.locale
       #   /Users/bricesanchez/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/sqlite3-1.3.13/lib/sqlite3/database.rb:91:in `initialize'
```

I had seen that i will need this method so i had keep this logic.

core/refinerycms-core.gemspec
```diff
@@ -23,7 +23,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version = Refinery::Version.required_ruby_version
 
   s.add_dependency 'refinerycms-i18n',            ['~> 4.0', '>= 4.0.0']
-  s.add_dependency 'awesome_nested_set',          ['~> 3.0', '>= 3.0.0']
```
parndt: nice, I guess core doesn't use this. Hopefully if extensions use it they specify it.

pages/lib/refinery/pages/finder.rb
shioyama: Ok, I figured out what's going on. The problem is that the original `conditions` hash has a `locale` key with an array of locales as its value. `translations_conditions` passes this on to the translations table, which uses it in an `IN` clause, but Mobility doesn't handle this one so you get an exception about a missing `locale` column on the parent model table.

This whole method can still be removed though, if you just update `with_mobility` to look like this:

```ruby
def with_mobility
  mobility_conditions = conditions.dup
  locales = mobility_conditions.delete(:locale)

  # A join implies readonly which we don't really want.
  Page.i18n.where(mobility_conditions).where(Page::Translation.table_name.to_sym => { locale: locales }).readonly(false)
end
def with_mobility
  mobility_conditions = conditions.dup
  locales = mobility_conditions.delete(:locale)

  # A join implies readonly which we don't really want.
  Page.i18n.where(mobility_conditions).
    where(Page::Translation.table_name.to_sym => { locale: locales }).readonly(false)
end
```

shioyama: Note that you don't need `joins(:translations)` here, Mobility will do that for you in the `i18n` scope.

shioyama: Actually wait, that doesn't quite work...

shioyama: Ok, long story short, this requires shioyama/mobility#51 to really work. The problem is that Mobility joins *ON* the current locale, and this clobbers this refinery query so you can't query on more than one locale. So for now, I'd say just leave this, and once shioyama/mobility#51 is done it should be refactorable.

shioyama: btw, good test coverage! Only 5 tests ended up failing, but they highlighted the problem here.

parndt: I'm glad! 🥇

shioyama:
Just an update: I put off releasing Mobility 0.6 because I want to implement #51, which would also allow us to remove that `translations_conditions` method here. Hoping to get that done this week or next. If you want to release with the `in_locale` method let me know and I can release 0.5.2 with just that in the meantime.

bricesanchez:
@shioyama we can wait after you, i already use this PR on a new project to see if everything is ok.

shioyama:
In shioyama/mobility#210 I've got working code that will now accept an optional locale or array of locales and query accordingly, like this:

```ruby
Post.i18n.where(titile: "foo", locale: [:en, :fr])
```

which would look for posts with a title of "foo" in either English or French. I believe with this change, you should be able to remove `translations_conditions`. Could you try with the branch in that PR?

If this looks good, I'll clean up the code and release 0.6 with this as a new feature.

Issue #3376, opened 2018-07-06
update dragonfly initializer to be compatible with dragonfly s3 datastore


kluucreations:
s3_datastore initialization is passing invalid opts into s3_data_store.
Update field names so that config values can be picked up succesfully

robbyrussell:
Hey there -- was this released in 4.0.3?

robbyrussell:
We ran into a problem upgrading a Rails 4 Refinery site to Rails 5 and bumping up Refinery in the process with the S3/Dragonfly configurations. We found the fix for this would resolve our config problem but wasn't in latest Rubygem releases. We're tracking 4-0-stable for now

parndt:
@robbyrussell curious.. maybe I didn't release it properly :sob: 

By the way thank you *so* much for Oh My Zsh!

parndt:
@robbyrussell it's released as refinerycms-dragonfly 1.0.1 - please let me know how you get on

PR #3377, opened 2018-07-06
Closes GH-3376

kluucreations:
looking at https://github.com/markevans/dragonfly-s3_data_store/blob/master/lib/dragonfly/s3_data_store.rb#L16, it seems that the opts passed in is expecting fields without the s3_ prefix.
this PR updates the opts passed in when initializing datastore to use the correct fields.


bricesanchez:
Hi @kluucreations! 

Thanks for your bugfix, do you think you could provide a test for this bugfix?

bricesanchez:
Thanks @kluucreations !

PR #3385, opened 2018-07-26
Interference Marketable Routes & ActiveStorage fixed

### LINKED ISSUE #3383: Interference between Marketable Pages & ActiveStorage (Rails 5.2)
RefineryCMS (with Rails 5.2. support) in combination with SpreeCommerce 3.6.2:

When mounting Refinery in root
```
Rails.application.routes.draw do
  mount Spree::Core::Engine, at: '/shop'
  mount Refinery::Core::Engine, at: '/'
end
```

marketable_page routes interfere with ActiveStorage, i.e. the Refinery route matches and  Refinery::PagesController takes over request:
```
Started GET "/rails/active_storage/disk/eyJf [...] g.jpeg" for 127.0.0.1 at 2018-07-20 21:31:58 +0200
Processing by Refinery::PagesController#show as JPEG
  Parameters: {"content_type"=>"image/jpeg", "disposition"=>"inline; filename=\"ror_bag.jpeg\"; filename*=UTF-8''ror_bag.jpeg", "path"=>"rails/active_storage/disk/eyJ [..] g", "locale"=>"en"}
```
I'm not really sure why this happens as the ActiveStorage routes are above the Refinery routes:
```
rake routes | grep 'pages#show\|rails'
[...]
rails_disk_service GET  /rails/active_storage/disk/:encoded_key/*filename(.:format)                              active_storage/disk#show
[...]
marketable_page GET    /*path(.:format)                           refinery/pages#show
[...]
```
Removing the mount point for Refinery or moving it to a different path however resolves the issue.

I resolve this now by a hacky route constraint in Refinery, however I would be interested why this happpens and how to resolve it 'right'.
```
def append_marketable_routes
  Refinery::Core::Engine.routes.append do
    get '*path', :to => 'pages#show', constraints: lambda { |req| !req.env["REQUEST_URI"].starts_with? "/rails/active_storage/" }, :as => :marketable_page
  end
end
```

Obi-TOB:
Closes #3383

### DIFF
--- a/pages/lib/refinery/pages/configuration.rb
+++ b/pages/lib/refinery/pages/configuration.rb
@@ -8,7 +8,7 @@ module Refinery
                     :cache_pages_full, :layout_template_whitelist,
                     :use_layout_templates, :page_title, :absolute_page_links, :types,
                     :auto_expand_admin_tree, :show_title_in_body,
-                    :friendly_id_reserved_words, :layout_templates_pattern, :view_templates_pattern,
+                    :friendly_id_reserved_words, :reserved_paths, :layout_templates_pattern, :view_templates_pattern,
                     :add_whitelist_elements, :add_whitelist_attributes, :whitelist_elements, :whitelist_attributes,
                     :home_page_path
 
@@ -64,6 +64,7 @@ module Refinery
     self.friendly_id_reserved_words = %w(
       index new session login logout users refinery admin images
     )
+    self.reserved_paths = %w(/rails/active_storage)
     self.layout_templates_pattern = 'app', 'views', '{layouts,refinery/layouts}', '*html*'
     self.view_templates_pattern = 'app', 'views', '{pages,refinery/pages}', '*html*'
   end
--- a/pages/lib/refinery/pages/configuration.rb
+++ b/pages/lib/refinery/pages/configuration.rb
@@ -64,7 +64,7 @@ module Refinery
     self.reserved_paths = %w(/rails/active_storage)
     self.friendly_id_reserved_words = %w(
       index new session login logout users refinery admin images
-    ) + self.reserved_paths.map { |path| path.split('/') }.flatten.uniq - [""]
+    ) + self.reserved_paths.map { |path| (path.split('/') - [""]).first}.flatten.uniq
     self.layout_templates_pattern = 'app', 'views', '{layouts,refinery/layouts}', '*html*'
     self.view_templates_pattern = 'app', 'views', '{pages,refinery/pages}', '*html*'
   end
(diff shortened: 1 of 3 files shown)


bricesanchez:
Thank you very much @Obi-TOB !

Should we also add `rails/active_storage`slug in `friendly_id_reserved_words` ?

We don't want to create a page with this slug and not able to reach it.

Obi-TOB:
Hi @bricesanchez,

I added all words from all reserved_paths to the friendly_id_reserved_words.

I guess this might go a bit far... as this probably means no friendly_id slugs with rails or active storage. I'm not entirely sure how friendly id stores the slugs, but I think we need to exclude each word on it's own.

Let me know what you think...

pages/lib/refinery/pages/configuration.rb
```diff
@@ -61,9 +61,10 @@ def layout_template_whitelist
     self.absolute_page_links = false
     self.types = Types.registered
     self.auto_expand_admin_tree = true
+    self.reserved_paths = %w(/rails/active_storage)
     self.friendly_id_reserved_words = %w(
       index new session login logout users refinery admin images
-    )
+    ) + self.reserved_paths.map { |path| path.split('/') }.flatten.uniq - [""]
```
bricesanchez: You probably just need to add `rails` because we already just add `refinery` or `admin` in `friendly_id_reserved_words ` even if there are sub path to these paths.

Obi-TOB: I now add only the first elements from each element of reserved_paths to friendly_id_reserved_words, so in the case we have only one element '/rails/active_storage' it will only take 'rails'. If you add several paths it will remove all 'root'-fragments of the paths

parndt: maybe we could use:

```ruby
    self.friendly_id_reserved_words = %w(
      index new session login logout users refinery admin images
    ) | reserved_paths.map { |path| path.split('/') }.flatten.uniq.reject(&:blank?)
```

Obi-TOB: I changed it to

```
self.friendly_id_reserved_words = %w(
  index new session login logout users refinery admin images
) | self.reserved_paths.map { |path| path.split('/').reject(&:blank?).first}.flatten.uniq
```
to take only the all root_fragments of the reserved paths.

bricesanchez:
Thanks @Obi-TOB !

I've just seen that the new config `reserved_paths` is missing pages initializer template : https://github.com/refinery/refinerycms/blob/master/pages/lib/generators/refinery/pages/templates/config/initializers/refinery/pages.rb.erb

Could you add it ?

bricesanchez:
Closed in favor of #3387.

PR #3390, opened 2018-08-20
Fix page tree cache refresh on sub page changes

aiomaster:
Hi,
it took me some hours to find and fix a problem that came in with admin page index collection caching (see https://github.com/refinery/refinerycms/commit/a1b6e8ac4831aef0747eb60223bf1e94df0ece19).
Problem was, that the page tree showed old pages if we changed a page title on a sub child of one of the children of a page.
The collection caching just checked the children of a page, but not if a child (or child of a child ...) was changed. So we ended with a wrong page tree.
I wrote a test and fixed the problem by including the descendants of a page into the caching decision.
Cheers!

### DIFF
--- a/pages/app/views/refinery/admin/pages/_page.html.erb
+++ b/pages/app/views/refinery/admin/pages/_page.html.erb
@@ -48,6 +48,6 @@
   </div>
 
   <%= content_tag :ul, class: 'nested', data: { 'ajax-content': refinery.admin_children_pages_path(page.nested_url) } do %>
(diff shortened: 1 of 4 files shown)


aiomaster:
I really don't know why the travis tests with mysql fail. I can not reproduce it on my local machine. Strange!

aiomaster:
Funny, it was a caching problem :)
added a `Rails.cache.clear` seemed to fix it.

bricesanchez:
@aiomaster thanks you so much, i did not test this feature on a sub child!

PR #3403, opened 2018-09-19
Now validate resources mime_types

bricesanchez:
* Accept only file types we allow
* Add ability to change the default configuration.
* Run and fix offenses with Rubocop


parndt:
I routinely upload all kinds of file types (including images) as resources, yet this defaults Refinery to only accepting PDFs. I'm not sure I'm happy with it being that restrictive..

bricesanchez:
which files types would you like as default?, i don't want to allow files that can be executed without permission.

parndt:
that's fair, however I think we should support all of the standard image formats that images does, text files, etc as well

PR #3405, opened 2018-10-10
Add Matomo open web analytics

brigriffin:
This PR adds the possibility for a site to use the Matomo open web analytics (a.k.a. Piwik) for their web analytics instead of or along with google analytics. It can be enabled in the core config by setting the two required parameters `config.matomo_analytics_server` and `config.matomo_analytics_site_id`. It works analog to the code for enabling google analytics and as such is pretty straightforward.

Please do let me know if there is anything you would like me to adapt or change.


bricesanchez:
Thanks @brigriffin for your contribution!

I think you should create an extension `refinerycms-matomo` to support this web analytics, we try to keep the core as lightweight as possible.

Do you agree @parndt ?

brigriffin:
@bricesanchez you're welcome, I'm happy to contribute to for such a great CMS.

Just wanted to add my opinion regarding the extension idea. I think an extension is a bit of an overkill: indeed there are only 27 lines of additional code spread over 4 already existing files... It would be an advantage for RefineryCMS to make it easy for people to chose for NOT going with google analytics and having an alternative choice out-of-the-box which takes in account privacy such as matomo. Nowadays people are much more privacy-aware...

parndt:
Yeah, I agree with promoting more privacy conscious options. I'm happy enough to have this in core @bricesanchez as we already have Google's one.

bricesanchez:
@brigriffin

>  I think an extension is a bit of an overkill

I agree if it's only for one service. Perhaps, the best should be to support more services in one extension, something like `refinerycms-analytics` with Matomo, Google Analytics,...

> having an alternative choice out-of-the-box which takes in account privacy such as matomo. 
 Nowadays people are much more privacy-aware...

I totally agree

> I'm happy enough to have this in core @bricesanchez as we already have Google's one.

If @parndt is agree, so I agree too !

brigriffin:
Thanks for merging and the interesting discussion around this topic. I am convinced this is a nice thing for RefineryCMS users being now able to chose between two different options in terms of web analytics, especially considering the fact that one of the options is relevant for privacy-aware people.

PR #3412, opened 2018-11-14
Replace redirect_to :back with redirect_back for Rails 5.2

victoryspirit:
In Rails 5.0 redirect_to :back is deprecated in favour of redirect_back [Add `redirect_back` and deprecate `redirect_to :back`](https://github.com/rails/rails/pull/22506)


bricesanchez:
Thanks @victoryspirit !

Issue #3411, opened 2018-11-06
Can we remove dragonfly-s3_data_store dependency?


veger:
When I install refinerycms-dragonfly I automatically pull in AWS S3 support (dragonfly-s3_data_store, and some other dependencies), even if I leave the configuration disabled and never going to use AWS S3 in my project.

Can we remove the (forced) dependency from the gem (and maybe put a a comment next to the [`self.s3_datastore`](https://github.com/refinery/refinerycms/blob/master/dragonfly/lib/refinery/dragonfly/configuration.rb#L61) to include this dependency when you put this setting to true?

bricesanchez:
Hi @veger !

Sure ! Could you provide a pull request ?

Thanks !

veger:
I was a bit busy and away for a while, but I finally made the PR

veger:
PR got merged :tada:

PR #3413, opened 2018-11-18
Make dragonfly-s3_data_store gem dependency optional

veger:
Remove forced dependencies for AWS S3 support (which (probably) makes no sense in a lot of projects)

See #3411


veger:
BTW I do not see a `EXTENSION=dragonfly` in travis... Is dragonfly actually being tested?! :open_mouth:

bricesanchez:
Thanks @veger!

Probably not if it's not in the travis config, could you add it please?

PR #3416, opened 2018-11-27
change capybara and rails-controller-testing dependencies to minimal versions and fix factory bot deprecation warnings

stevereinke:
Hi there, 
we are using refinerycms in our project and tried to use refinerycms-testing. We got some problems with the development dependencies of the testing gem and our dev dependencies. So we tried to use new versions of the gems mentioned above and it just worked. So is ti possible to update the dev dependencies for refinerycms-testing?
Thanks,
Steve

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,7 @@ group :test do
   gem 'launchy'
   gem 'coveralls', require: false
   gem 'rspec-retry'
+  gem 'puma'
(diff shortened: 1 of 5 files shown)


parndt:
Thanks @stevereinke :+1:

stevereinke:
Thanks for Merge.

PR #3418, opened 2018-12-20
[Admin UI] tweak litter style and use the more beautiful rich editor.

lanzhiheng:


lanzhiheng:
@bricesanchez I think we need a more beautiful rich editor for New Admin UI.

lanzhiheng:
@bricesanchez Thank you for your comments. I will try to search the suitable editor for it.

lanzhiheng:
@bricesanchez I have change the editor like below.  I need to add an extra api for image upload I am not sure if you can accept it. please give me some feedback. 

image

I just include very basic feature for ckeditor, if you need powerful function, I can rebuild a new package with some plugin.

core/app/assets/javascripts/refinery/refinery.js.erb
```diff
@@ -10,4 +10,6 @@
  *= require jquery/jquery.browser
  *= require admin
  *= require refinery/core
+ *= require ckeditor
```
parndt: where is this dependency defined?

parndt:
@bricesanchez what do you think of the above? I think the UI looks nice, but probably it should be an extension.

lanzhiheng:
@parndt CKeditor have the licence issue. So I create a special engine base by trix which have the MIT licence https://github.com/lanzhiheng/refinerycms-trix . It is very easy to extend. But the problem is that the editor is too poor https://github.com/basecamp/trix/issues/587 .

parndt:
Will all this be achieved in Rails 6 by https://github.com/rails/rails/tree/master/actiontext ?

bricesanchez:
@parndt yes it should be an extension, i think we could close this PR.

PR #3425, opened 2019-02-12
Suggest ruby-install and chruby

parndt:
I prefer these options, and have used them for years.


bricesanchez:
Now i use ASDF https://github.com/asdf-vm/asdf

So i don't know what to answer :D

parndt:
I'm inclined to just leave it alone

PR #3436, opened 2019-08-28
Get S3 credentials from the encrypted credentials file instead of ENV

anitagraham:
With the introduction of `config/credentials.yml.enc` we can avoid using ENV to store what should be secure keys.

Notes:

- `bucket_name` and `region` aren't strictly security keys, but they seems to belong with the other keys here
-  We could continue to use ENV  as an alternative to credentials. However once people have to create a credentials or secrets file, they may as well include their s3 keys in it.


parndt:
I was going to ask about existing apps but then I realised that we're just changing the generated initialiser, so the upgrade path is strictly opt-in!

PR #3442, opened 2019-10-01
Use collection caching for generation of page links (pages_dialogs) 

evenreven:
Using the link button in wymeditor scales poorly when you get many pages. It was okay in the beginning, but now my app has approx 700 pages (most of them in two languages), and loading the page_links partial takes 15 seconds or more in my application. (This is especially annoying since most of our links are external and thus the page_links aren't even necessary most of the time.) Since the partial seems to generate the page link list from scratch every time and uses the standard page model as collection to create the links, it's a prime candidate for collection caching (since invalidation is taken care of by the model). 

I've done some simple testing. Load time went from approx. 15 seconds to approx 1 second. I've created a new page and I've reordered pages. No stale content. Link list is significantly faster even when there are new entries. (After reordering it's quite slow on first load, but still faster than uncached.)

I've filed this PR to the 4-0-stable branch since that's the branch I use (because I use page-images, which is still on Globalize, so I can't go to master (also, I have a lot of Globalize-specific code in my app)). Please consider merging anyway, or if you want to merge this to master, please backport/cherry-pick this to the 4-0-stable branch as well if it's not too much trouble. Thanks!

### DIFF
--- a/pages/app/views/refinery/admin/pages_dialogs/link_to.html.erb
+++ b/pages/app/views/refinery/admin/pages_dialogs/link_to.html.erb
@@ -27,7 +27,7 @@
       <div id="your_page_content" class="clearfix">
         <div id="pages_list" class="pages_list">
           <ul class="link_list">
-            <%= render :partial => 'page_link', :collection => @pages, :cached => 'true',
+            <%= render :partial => 'page_link', :collection => @pages, :cached => true,
                        :locals => {
                          :child => 0,
                          :link_to_arguments => {}
--- a/pages/app/views/refinery/admin/pages_dialogs/link_to.html.erb
+++ b/pages/app/views/refinery/admin/pages_dialogs/link_to.html.erb
@@ -27,7 +27,7 @@
       <div id="your_page_content" class="clearfix">
         <div id="pages_list" class="pages_list">
           <ul class="link_list">
-            <%= render :partial => 'page_link', :collection => @pages,
+            <%= render :partial => 'page_link', :collection => @pages, :cached => 'true',
                        :locals => {
                          :child => 0,
                          :link_to_arguments => {}


bricesanchez:
Hi @evenreven ! Could you rebase your branch with master?

evenreven:
Since this PR is for 4-0-stable (and from my local branch called the same), I'm not sure how to do that easily.

bricesanchez:
Oh sorry i did not see it was for 4-0-stable. CI for this branch does not works anymore. I will try it on master before merging it in this branch

evenreven:
@bricesanchez I don't mean to nag you about this, but is there a realistic chance of this getting merged in the near future? I'm asking because, if not, I'll start using my fork of the 4-0-stable branch with this change in production instead. I absolutely understand if you don't have the time to review and merge, I just want to avoid forking if you're planning to look at this anyway.

We create several links every day, and the 15 seconds waiting time on each one is starting to drive both me and my colleagues nuts. :)

pages/app/views/refinery/admin/pages_dialogs/link_to.html.erb
```diff
@@ -27,7 +27,7 @@
       <div id="your_page_content" class="clearfix">
         <div id="pages_list" class="pages_list">
           <ul class="link_list">
-            <%= render :partial => 'page_link', :collection => @pages,
+            <%= render :partial => 'page_link', :collection => @pages, :cached => 'true',
```
bricesanchez:
```suggestion
            <%= render :partial => 'page_link', :collection => @pages, :cached => true,
```

bricesanchez:
Sorry for the delay @evenreven, Specs are ok on master with this feature #3459 so we can merge it when you could make the small change on the boolean.

evenreven:
That's weird! How did that even work? I'm not used to the old Ruby syntax, so I was afraid I'd screwed something up, thanks for checking. Fixed now.

PR #3451, opened 2019-10-31
Fix the configuration for Dragonfly so you can use the before_s…

joebutler2:
This fixes an issue with setting the `Dragonfly` configuration for `before_serve`. It looked like before it was a simple attr_accessor, but since `before_serve` is more like a DSL it was expecting to be called with a block. This PR corrects that, and I'm able to confirm it works on my project.

The good news is this also solves another use case, serving private content. If you look at this Google Groups post from a few years ago, it appears my client isn't alone in wanting this feature. https://groups.google.com/forum/#!searchin/refinery-cms/dragonfly$20before_serve%7Csort:date/refinery-cms/lRBOsddSzHM/1tXNjzaHcBUJ

### DIFF
--- a/dragonfly/lib/refinery/dragonfly/dragonfly.rb
+++ b/dragonfly/lib/refinery/dragonfly/dragonfly.rb
@@ -32,7 +32,7 @@ module Refinery
 
           #  These options require a name and block
           define_url extension.dragonfly_define_url if extension.dragonfly_define_url.present?
-          before_serve extension.dragonfly_before_serve if extension.dragonfly_before_serve.present?
+          before_serve(&extension.dragonfly_before_serve) if extension.dragonfly_before_serve.present?
 
 
           # There can be more than one instance of each of these options.
--- a/dragonfly/refinerycms-dragonfly.gemspec
+++ b/dragonfly/refinerycms-dragonfly.gemspec
@@ -3,7 +3,7 @@
 Gem::Specification.new do |s|
   s.platform          = Gem::Platform::RUBY
   s.name              = %q{refinerycms-dragonfly}
-  s.version           = '1.0.1'
+  s.version           = '1.0.2'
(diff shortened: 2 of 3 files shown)


joebutler2:
The build is failing due to the upgrade to Rails 6. Some of the functionality of this gem requires us to use the relatively new credentials feature.

anitagraham:
Must be all the rage. I just submitted a similar PR as the same problem was reported on the refinerycms google group. I'll pull my PR.

dragonfly/lib/refinery/dragonfly/dragonfly.rb
```diff
@@ -32,7 +32,7 @@ def configure!(extension)
 
           #  These options require a name and block
           define_url extension.dragonfly_define_url if extension.dragonfly_define_url.present?
-          before_serve extension.dragonfly_before_serve if extension.dragonfly_before_serve.present?
+          before_serve(&extension.dragonfly_before_serve) if extension.dragonfly_before_serve.present?
```
parndt: makes sense, otherwise it'd get immediately executed

parndt:
I've fixed the build failure in #3453

dragonfly/lib/refinery/dragonfly/dragonfly.rb
joebutler2: Since you're passing in a block, it would be a `Proc` object at this point. So it would only execute if you invoked the `call` method on it. The `&` is converting the `Proc` object back into a block since the `before_serve` method is expecting a block.

I'm pretty sure that is the case, let me know if I mistook something.

joebutler2:
@parndt is this good to merge in? Do you want me to do anything else?

parndt:
@joebutler2 I would just like a review from @anitagraham 😄

PR #3452, opened 2019-11-06
Dragonfly before serve

anitagraham:
I think this will fix a problem reported in  the [refinery-cms google group.](https://groups.google.com/forum/#!topic/refinery-cms/lRBOsddSzHM).

Usage of dragonfly before_save option, from Refinery:

```Ruby
#config/initializers/refinery/dragonfly.rb

Refinery::Dragonfly.configure do |config|
...
  bar =  -> (job, env) {
    throw :halt, [403, {'Content-Type' => 'text/plain'}, ["You shall not pass"]]
  }
  config.before_serve = bar
...
end
```


parndt:
#3451

PR #3459, opened 2019-11-26
Add collection caching page links

bricesanchez:
Change proposed by @evenreven


evenreven:
Thank you for reviewing and merging this @bricesanchez and @parndt ! I have around 700 pages in my project, most of them bilingual (if that matters in this case). I just deployed this change to production now, and clicking insert link went from 12 seconds down to 0.13 seconds! 

So hopefully this helps other users too. I have created pages and tried insert links again, and I see no issues with stale cache, but I'll keep my eyes peeled in the days and weeks to come and will report back if I come across anything. But using collection caching on a standard model like this should be safe.

bricesanchez:
Thanks for your help @evenreven!

PR #3463, opened 2020-01-09
Bugfix/wait for ajax

bricesanchez:


parndt:
Hmmm @bricesanchez interesting. I wonder if the webdriver itself has a config option for how long to wait. For example the `cuprite` runner has a [slowmo option](https://github.com/machinio/cuprite#running-in-docker) 

> :slowmo (Integer | Float) - Set a delay to wait before sending command. Useful companion of headless option, so that you have time to see changes.

We don't use cuprite, but it's an interesting idea.

PR #3464, opened 2020-01-13
Bugfix/zeitwerk

bricesanchez:


Gemfile
```diff
@@ -3,10 +3,10 @@ source 'https://rubygems.org'
 gemspec
 
 path "./" do
-  gem "refinerycms-core"
-  gem "refinerycms-images"
-  gem "refinerycms-pages"
-  gem "refinerycms-resources"
+  gem "refinery_core"
```
parndt: 🙅‍♂

bricesanchez: Yes, i'm also not ok for this change, it will be a pain. I'm trying to find a better way to fix Zeitwerk autoloader

PR #3465, opened 2020-01-13
Comply with Zeitwerk autoloader

parndt:
### Context

For full Zeitwerk support, we require all gems that currently depend on
something in Refinery by requiring the gem's name to update the require
to a path inside the gem.

For example, if a gem had:

    require 'refinerycms-pages'

This has to change to:

    require 'refinery/pages'

Not a huge code change, but needs to be done everywhere.

### Changes

Applies the above suggestion to code that lives inside this repository.

### Considerations

One issue is that this kind of violates the namespace expectation of a 
gem called `refinerycms-core`, which should have `lib/refinerycms/core.rb`.

Perhaps this is worth a lot of consideration as an alternative.

Resolves #3464


parndt:
One niggly issue is that this kinda violates the namespace expectation of a gem called `refinerycms-pages`, which should have `lib/refinerycms/pages.rb`.

Perhaps this is worth a lot of consideration as an alternative.

core/lib/refinery/cli.rb
```diff
@@ -1,7 +1,7 @@
 require 'thor'
 
 module Refinery
-  class CLI < Thor
```
parndt: perhaps it should live inside `Core` too?

bricesanchez: I think we could leave it here until we rewrite it

Gemfile
```diff
@@ -15,10 +15,10 @@ gem 'listen'
 gem 'activejob'
 
 # Add support for refinerycms-acts-as-indexed
-gem 'refinerycms-acts-as-indexed', ['~> 3.0', '>= 3.0.0']
+gem 'refinerycms-acts-as-indexed', ['~> 3.0', '>= 3.0.0'], require: 'refinery/acts_as_indexed'
```
parndt: still wishing there is a way we can get this without the `require`. In this case I guess the gem should be called `refinerycms-acts_as_indexed`

bricesanchez: It could help but there will always be a problem on the `cms` part

parndt: not if it's `lib/refinerycms/acts_as_indexed.rb`

Maybe for now we can use `lib/refinerycms/acts/as/indexed.rb`? Haven't tried it yet

parndt:
@bricesanchez not exactly a monkey patch, we're just defining our own inflector right?

Gemfile
```diff
@@ -10,11 +10,15 @@ path "./" do
 end
 
 # routing-filter needs a newer release than version 0.6.3
-gem 'refinerycms-i18n', github: 'refinery/refinerycms-i18n', branch: 'master'
+gem 'refinerycms-i18n', github: 'refinery/refinerycms-i18n', branch: 'zeitwerk'
 gem 'routing-filter', github: 'svenfuchs/routing-filter', branch: 'master'
 
+gem 'activejob'
+
```
parndt, line 17: do we need this?

RandieM:
This seems to be almost done. It would be nice to see it merged.

parndt:
Does it work for you @RandieM ?

RandieM:
> Does it work for you @RandieM ?

Sorry for the late reply. The error is gone and the website seems to work without issues. Let me know if you'd like me to test something more specific.

Gemfile
```diff
@@ -9,10 +9,14 @@ path "./" do
   gem "refinerycms-resources"
 end
 
-gem 'refinerycms-i18n', github: 'refinery/refinerycms-i18n', branch: 'master'
+gem 'refinerycms-i18n', github: 'refinery/refinerycms-i18n', branch: 'zeitwerk'
+
+gem 'activejob'
```
parndt, line 14:
```suggestion
```

parndt:
@RandieM thanks 😄

bricesanchez (approved):

resources/lib/refinery/resources.rb
```diff
@@ -1,5 +1,5 @@
-require 'refinerycms-core'
-require 'refinerycms-dragonfly'
+require 'refinerycms/core'
+require 'refinerycms/dragonfly'
```
dgm, line 2: Shouldn't this be 'refinery/dragonfly'   then?  I still get " cannot load such file -- refinerycms/dragonfly (LoadError)"  as I try to upgrade to rails 6

PR #3487, opened 2021-01-08
Configure S3 access using Rails:credentials 

anitagraham:
I have update the documentation on using S3 for uploading to include using Rails credentials.


anitagraham:
Sorry. My troubled relationship with Git continues unabated. I though I was just pushing a single change from a single branch.

PR #3490, opened 2021-03-06
Remove call to URI.Encode for dragonfly.cache_store_root.

anitagraham:
URI.Encode is deprecated.
The line is commented 
```
URI.encode("file:#{extension.dragonfly_cache_store_root}/meta"), # URI encoded in case of spaces
```

This variable being encoded is used as a file path and it  is not clear that encoding spaces is useful in this case.


parndt (approved):
seems reasonable

PR #3492, opened 2021-03-06
Refinery won't work with Rails 6.1 until 'routing-filter' is updated.…

anitagraham:
… Change comes from  PR https://github.com/rails/rails/pull/39534, which converts the route params array into an object.

### DIFF
--- a/core/refinerycms-core.gemspec
+++ b/core/refinerycms-core.gemspec
@@ -3,7 +3,7 @@
 require File.expand_path('../core/lib/refinery/version', __dir__)
 
 version = Refinery::Version.to_s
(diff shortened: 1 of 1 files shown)


parndt:
@anitagraham you can work around this by using the routing-filter from GitHub:

https://github.com/refinery/refinerycms/blob/6a1520e166a73d15ea14afc6fc3ba8d08f33e210/Gemfile#L12-L14

@simi hey Josef 👋 ❤️, do you know if there's an update to `routing-filter` expected?

parndt:
Looks like version 0.7.0 is out

Issue #3523, opened 2023-01-31
`superclass mismatch` on Rails/Zeitwerk reload


richard-kramer:
While upgrading an existing Rails app to `v6.1` and enabling the zeitwerk autloader, I stumbled across this error, when reloading the application in development mode:

> `superclass mismatch for class FinderByTitle (TypeError)`

This happens everywhere, code was reloaded.

I tried to reproduce the problem in a minimal setup by initializing a new rails 6.1 app, including `refinerycms` in the Gemfile: [`webit-de/refinerycms-zeitwerk-reloading-reproduction`](https://github.com/webit-de/refinerycms-zeitwerk-reloading-reproduction)

```sh
echo 'Refinery::Page; reload!; Refinery::Page' | rails c
# >
# Running via Spring preloader in process 408055
# Loading development environment (Rails 6.1.7.2)
# Switch to inspect mode.
# Refinery::Page; reload!; Refinery::Page
# Reloading...
# ~/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/refinerycms-aee49a603860/pages/lib/refinery/pages/finder.rb:62:in `<module:Pages>': superclass mismatch for class FinderByTitle (TypeError)
```

---

The issue stems from the definition of `Refinery::Pages::Finder` and its sibling classes (`Refinery::Pages::FinderByTitle` and so on). These are all defined beside each other in [`pages/lib/refinery/pages/finder.rb`](https://github.com/refinery/refinerycms/blob/aee49a603860bf7d5fdb1532b9add5f4e7f839f4/pages/lib/refinery/pages/finder.rb) which gets required in [`pages/app/models/refinery/page.rb`](https://github.com/refinery/refinerycms/blob/aee49a603860bf7d5fdb1532b9add5f4e7f839f4/pages/app/models/refinery/page.rb#L6).

When reloading, zeitwerk reloads the model, which in turn loads the finder again. But because the file is named `finder.rb`, only `Refinery::Pages::Finder` gets unloaded and reloaded again. All other classes stay loaded and would be patched by the reload. But because `Refinery::Pages::Finder` has been reloaded and initialized again, it got a new `object_id`, thus triggering the **`superclass mismatch`** error.

One solution, that fixed the error, was to move all sibling-classes of `Refinery::Pages::Finder` into it, making them children of `Refinery::Pages::Finder`. Because they are only called inside `Refinery::Pages::Finder` itself, this *should* not make problems.

[...]

PR #3524, opened 2023-01-31
Fix Zeitwerk conventions for `Refinery::Pages::Finder`

richard-kramer:
Regarding #3523

The issue stems from the definition of `Refinery::Pages::Finder` and its sibling classes (`Refinery::Pages::FinderByTitle` and so on). These are all defined beside each other in [`pages/lib/refinery/pages/finder.rb`](https://github.com/refinery/refinerycms/blob/aee49a603860bf7d5fdb1532b9add5f4e7f839f4/pages/lib/refinery/pages/finder.rb) which gets required in [`pages/app/models/refinery/page.rb`](https://github.com/refinery/refinerycms/blob/aee49a603860bf7d5fdb1532b9add5f4e7f839f4/pages/app/models/refinery/page.rb#L6).

When reloading, zeitwerk reloads the model, which in turn loads the finder again. But because the file is named `finder.rb`, only `Refinery::Pages::Finder` gets unloaded and reloaded again. All other classes stay loaded and would be patched by the reload. But because `Refinery::Pages::Finder` has been reloaded and initialized again, it got a new `object_id`, thus triggering the **`superclass mismatch`** error.

The solution, that fixed the error, was to move all sibling-classes of `Refinery::Pages::Finder` into it, making them children of `Refinery::Pages::Finder`. Because they are only called inside `Refinery::Pages::Finder` itself, this *should* not make problems.

I am not sure, if this is the best or most elegant solution, if it introduces side effects or if there are other parts of refinery, that suffer the same issues.

### DIFF
--- a/pages/lib/refinery/pages/finder.rb
+++ b/pages/lib/refinery/pages/finder.rb
@@ -57,123 +57,122 @@ module Refinery
         translations_conditions
       end
 
-    end
-
-    class FinderByTitle < Finder
-      def initialize(title)
-        @title = title
-        @conditions = default_conditions
-      end
-
-      def default_conditions
-        { :title => title }
-      end
-
-      private
-      attr_accessor :title
-    end
-
-    class FinderBySlug < Finder
-      def initialize(slug, conditions)
-        @slug = slug
-        @conditions = default_conditions.merge(conditions)
-      end
-
-      def default_conditions
-        {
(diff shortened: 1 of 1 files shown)


parndt (approved):

parndt:
Thank you! 💜💛

